PR #5, opened 2022-07-20
Cordova plugins branch

barakbenhur1:
Cordova Pulgin
*add java script file that exposes the Optimobile Sdk functionality
*cerate a platform folder with 2 sub folders ios and android
*create a ios qa app


optimove-andy:
We should check in with Kosta how the QA apps usually work, they don't generally have a QA app in the SDK repo in my experience.

Possibly we extract this QA app and put it into azure.

PR #6, opened 2022-07-25
Project scaffolding

TomBarzilay1:
Initial commit + Add Initialization functionality + created a QA application that consumes the plugin logic


Cordova_QA_App/platforms/ios/Podfile
```diff
@@ -0,0 +1,9 @@
+# DO NOT MODIFY -- auto-generated by Apache Cordova
+source 'https://github.com/CocoaPods/Specs.git'
+platform :ios, '11.0'
+
+target 'C:\Users\tom_b\Desktop\Work repositories\Optimove-SDK-Cordova\Cordova_QA_App\platforms\ios\Cordova_QA_App' do
```
optimove-andy, line 5: Can we get rid of absolute paths.

Cordova_QA_App/optimove.json
```diff
@@ -0,0 +1,4 @@
+{
+ 	 "optimove_credentials": "WzEsIm15LXRva2VuIiwibXktY29uZmlnLWZpbGUiXQ==",
+	"optimove_mobile_credentials": "WzEsImV1LWNlbnRyYWwtMSIsIjE5ZGY1ZTc3LTJhNmMtNGIzZC05YmI5LWNiNTQ0MDgwMmEyNiIsIlMxTksyWjBMSVpWeVl6eVk4bnB0RlAzQjBpSDVjbU50RU5pSSJd"
```
optimove-andy, line 3: We shouldnt ship credentials - possibly worth checking how QA apps for Android / IOS bake these in.

plugins/www/Optimobile.js
```diff
@@ -0,0 +1,8 @@
+
+
+var exec = require('cordova/exec');
+
+
+window.initialize = function (arg0, arg1, success, error) {
```
optimove-andy, line 6: Can we have better arg names, assume arg0 / arg1 are the sets of credentials?

plugins/src/ios/OptimobileSDKPlugin.swift
```diff
@@ -0,0 +1,9 @@
+import OptimoveSDK
+@objc(Optimobile_Cordova) class OptimobileSwift : CDVPlugin {
+    @objc(initialize:)
+    func initialize(command: CDVInvokedUrlCommand) {
+        let args = command.arguments
+        let config = OptimoveConfigBuilder(optimoveCredentials: args?[0] as? String, optimobileCredentials: args?[1] as? String).enableInAppMessaging(inAppConsentStrategy: .explicitByUser).build()
```
optimove-andy, line 6: We shouldn't enable In-App messaging automatically, not all customers will take that feature.

optimove-andy:
We should check in with Kosta how the QA apps usually work, they don't generally have a QA app in the SDK repo in my experience.

Possibly we extract this QA app and put it into azure.

plugins/www/Optimobile.js
TomBarzilay1: the args name were changed to  arg0->optimove_credentials and arg1->optimove_mobile_credentials

Cordova_QA_App/optimove.json
TomBarzilay1: the json file was changed  to {
 	 "optimove_credentials": "YOUR_OPTIMOVE_CREDENTIALS",
	"optimove_mobile_credentials": "YOUR_OPTIMOVE_MOBILE_CREDENTIALS"
}

plugins/src/ios/OptimobileSDKPlugin.swift
TomBarzilay1: we added a parameter inAppConsentStrategy to determine if the Customer want to add In app messages feature

Cordova_QA_App/platforms/ios/Podfile
```diff
@@ -0,0 +1,9 @@
...
+	project 'C:\Users\tom_b\Desktop\Work repositories\Optimove-SDK-Cordova\Cordova_QA_App\platforms\ios\Cordova_QA_App.xcodeproj'
+	pod 'OptimoveCore', '5.0.0'
+	pod 'OptimoveSDK', '5.0.0'
```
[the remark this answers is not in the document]
TomBarzilay1: The SDK version was changed

Cordova_QA_App/platforms/ios/Podfile
TomBarzilay1: yes.

Cordova_QA_App/optimove.json
cgwyllie: Do we need to consider scrubbing these from git history? Maybe trickier now it's on the PR diffs.

plugins/www/Optimobile.js
cgwyllie: Point on naming, can we work with `lowerCamelCase` in general for JS/TS files?

Example_App/.gitignore
```diff
@@ -0,0 +1,25 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+.DS_Store
+
+# Generated by package manager
+node_modules/
+
+# Generated by Cordova
+
```
cgwyllie, line 25: I'm wondering if we should treat the `platforms` and `plugins` folders in the example app as 'generated' code. They should (theoretically) be re-added by `cordova prepare`, provided that the platforms and plugins are saved into the `config.xml` for the project.

It would greatly simplify the git structure and reduce churn for generated build artefacts etc.

What do you think?

cgwyllie: (perhaps if there's anything we want to customise like app icon this won't be so simple. Although IIRC from a past life as a Cordova application developer, we used some build step to cut & copy the assets into the correct folders at build time)

plugins/src/android/OptimobileInitProvider.java
```diff
@@ -0,0 +1,77 @@
...
+import android.net.Uri;
+import android.text.TextUtils;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.optimove.android.Optimove;
+import com.optimove.android.OptimoveConfig;
+
+public class OptimobileInitProvider extends ContentProvider {
```
cgwyllie, line 17: Naming: can we go with Optimove instead of Optimobile?

plugins/src/ios/OptimobileSDKPlugin.swift
cgwyllie: This should likely come from the configuration JSON file instead of an argument, but we can handle it in the follow-up.

plugins/plugin.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version='1.0' encoding='utf-8'?>
+<plugin id="com-optimobilecreate-plugins" version="1.0.0" xmlns="http://apache.org/cordova/ns/plugins/1.0" xmlns:android="http://schemas.android.com/apk/res/android">
```
cgwyllie, line 2: Plugin name / ID should be something like `optimove-cordova-plugin`, this will be the public name consumers use to install/reference the plugin.

plugins/plugin.xml
```diff
@@ -0,0 +1,40 @@
...
+    <name>OptimobileSDKPlugin</name>
+    <js-module name="Optimobile" src="www/Optimobile.js">
```
cgwyllie, line 4: +1 on naming Optimobile -> Optimove

plugins/scripts/Optimobile.plist
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cgwyllie, line 1: This file should more mimic the credentials file of https://github.com/Kumulos/KumulosSdkCordova/blob/master/scripts/kumulos.plist

cgwyllie: (can be handled in the second PR)

cgwyllie (approved):

PR #9, opened 2022-07-28
Cordova swift branch scaffolding

barakbenhur1:
Add Ios Changes


plugins/scripts/optimove.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
+        <key>apiKey</key>
+        <string>$(API_KEY)</string>
```
cgwyllie, line 6: I think these values should be:

1) updated to be the optimove & mobile credentials (two keys)
2) the placeholder syntax that `renderTemplate` uses in inject-config (e.g. `{{OPTIMOVE_CREDENTIALS}}`)

plugins/scripts/inject-config.js
```diff
@@ -83,8 +83,44 @@ module.exports = function injectOptimoveConfig(context) {
...
+
+    if (!fs.existsSync(configDest)) {
+        console.error(
+            'optimove: optimove.plist resource not found, aborting setup'
+        );
+        return;
+    }
+
+    const config = renderTemplate('optimove.plist', {
+        API_KEY: kumulosConfig.apiKey,
```
cgwyllie, line 115: OK think this looks good, just need to update the values to match the expected optimove / mobile credentials keys expected by the SDK for initialisation

plugins/scripts/optimove.plist
barakbenhur1: i can't compile to app with this syntax, it must be a valid plist

plugins/scripts/inject-config.js
barakbenhur1: Done

plugins/scripts/optimove.plist
cgwyllie: Before the app is compiled, the inject-config script should run and translate the syntax by replacing the `{{}}` tokens with real values, making the syntax correct.

The script can be triggered with `cordova prepare` in the test app and the output should give some info about running the hook.

So, in our plugin sources, the plist should be a template (and it won't be valid), but in the app's Resources bundle, it should be a valid one because it was processed by the script.

barakbenhur1: done

cgwyllie:
Closing in favour of #11

PR #11, opened 2022-07-28
Cordova swift branch initialize from config

barakbenhur1:
change initialize to read from config


plugins/scripts/inject-config.js
```diff
@@ -83,8 +83,44 @@ module.exports = function injectOptimoveConfig(context) {
...
+            'optimove: optimove.plist resource not found, aborting setup'
+        );
+        return;
+    }
+
+    const config = renderTemplate('optimove.plist', {
+        API_KEY: OptimoveConfig.apiKey,
+        SECRET_KEY: OptimoveConfig.secretKey,
+        ENABLE_CRASH: OptimoveConfig.enableCrashReporting,
+        IN_APP_STRATEGY: OptimoveConfig.inAppConsentStrategy
```
cgwyllie, line 118: I think we need to align the keys in the template plist, the inject-config script, and the swift code?

There should be two plist keys, one for optimove credentials, and another for mobile credentials, similar to Android.

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,51 @@
...
-      return.notEnabled
-    }()
-
-    let config = OptimoveConfigBuilder(optimoveCredentials: args?[0] as? String, optimoveCredentials: args?[1] as? String).enableInAppMessaging(inAppConsentStrategy: inAppConsentStrategy).build()
-    Optimove.initialize(with: config)
-  }
\ No newline at end of file
+import OptimoveSDK
+import NotificationCenter
+@objc(Optimobile_Cordova) class OptimobileSDKPlugin : CDVPlugin {
```
cgwyllie, line 3: Naming: can we use Optimove instead of Optimobile?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,51 @@
...
+    private let internalQueue = DispatchQueue(label: "com.singletioninternal.queue",
```
cgwyllie, line 4: Is this used?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,51 @@
...
+                                              qos: .default,
+                                              attributes: .concurrent)
+    
+    private var pendingPush: PushNotification?
+    
+    private let optimoveCredentialsKey = "optimoveCredentials"
+    private let optimoveMobileCredentialsKey = "optimoveMobileCredentials"
+    
+    private let credentialsJsonName = "optimove"
```
cgwyllie, line 13: Unused?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,51 @@
...
+        }
+        
+        let config = OptimoveConfigBuilder(optimoveCredentials: configValues[optimoveCredentialsKey], optimobileCredentials: configValues[optimoveMobileCredentialsKey])
+        
+        if configValues["inAppConsentStrategy"] == "auto-enroll" {
+            config.enableInAppMessaging(inAppConsentStrategy: .autoEnroll)
+        }
+        else if configValues["inAppConsentStrategy"] == "explicit-by-user" {
+            config.enableInAppMessaging(inAppConsentStrategy: .explicitByUser)
+        }
```
cgwyllie, line 47: Can we defer these pieces to the messaging branches?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,51 @@
...
-  }
\ No newline at end of file
+import OptimoveSDK
+import NotificationCenter
+@objc(Optimobile_Cordova) class OptimobileSDKPlugin : CDVPlugin {
+    private let internalQueue = DispatchQueue(label: "com.singletioninternal.queue",
+                                              qos: .default,
+                                              attributes: .concurrent)
+    
+    private var pendingPush: PushNotification?
```
cgwyllie, line 8: Can we defer these pieces to the messaging branches?

plugins/.gitignore
```diff
@@ -23,5 +23,9 @@
 node_modules/
 src/core/**/*.js
 npm-debug.log
+xcuserdata
+/platforms/
+/plugins/
+optimove.json
```
cgwyllie, line 29: Do these need adding to the plugin ignore? I'm not 100% sure, I think this already exists at the root repository level, and there are no Xcode projects inside the plugins dir itself.

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: yes this is for the thread safe init

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: yes

plugins/.gitignore
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: sure i have one open so when i make a pr from it will check that piece of code

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: done

plugins/scripts/inject-config.js
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: Ah yes, sorry had overlooked this.

Perhaps we could use something like the lazy property approach mentioned here https://stackoverflow.com/a/37887068 to avoid creating a queue just for initialisation? (i.e. we could de-ref some lazy prop with a block initialiser from the load method?)

plugins/scripts/optimove.plist
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>apiKey</key>
+	<string>{{API_KEY}}</string>
```
cgwyllie, line 6: I think we still need to align the plist key/values to match the inject-config template rendering?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,14 +1,49 @@
...
+        }
+        
+//        if let userInfo = notification.userInfo {
+//            let userInfoDict = notification.userInfo?[UIApplication.LaunchOptionsKey.remoteNotification];
+//            if userInfoDict != nil {
+//                pendingPush = PushNotification.init(userInfo: userInfo, response: nil)
+//            }
+//        }
+        
+//        let config = OptimoveConfigBuilder(optimoveCredentials: configValues[optimoveCredentialsKey], optimobileCredentials: configValues[optimoveMobileCredentialsKey])
```
cgwyllie, line 38: We should still have the setup of config from the plist values in here?

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: Think we maybe missed a ref to `OptimobileSDKPlugin`

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: you're right:)

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: I didn't understand that comment

plugins/scripts/optimove.plist
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: i think what i did is better

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: Ah sorry I just meant the class name is still Optimobile, can be renamed too :)

plugins/scripts/optimove.plist
```diff
@@ -0,0 +1,14 @@
...
+<plist version="1.0">
+<dict>
+	<key>optimoveCredentialsKey</key>
+	<string>{{optimoveCredentialsKey}}</string>
+	<key>optimoveMobileCredentialsKey</key>
+	<string>{{optimoveMobileCredentialsKey}}</string>
+	<key>enableCrashReporting</key>
+	<{{ENABLE_CRASH}}/>
+	<key>inAppConsentStrategy</key>
+	<string>{{IN_APP_STRATEGY}}</string>
```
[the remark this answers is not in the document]
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: OK, can you explain a little more why you think it's better?

My suggestion is based on the notes around excessive thread creation from https://developer.apple.com/documentation/dispatch/dispatchqueue

It seems perhaps we don't need to create a `DispatchQueue` to handle this if we use lazy properties. Because this queue would only be used for the initialisation, it seems a little wasteful.

Happy to hear your thoughts on it and comparison of approaches.

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: done

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: because this is the way to do thread safety in swift
lazy is for parameters as far as i know, because there is no need to make function initialize only when first used
it is a function:)

but i will have a look at what you send
i guess you mean that the config should be a lazy global parameter that get called first time in didFinishLaunching

barakbenhur1: make config init thread safe by making it lazy property of class

plugins/scripts/optimove.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>optimoveCredentialsKey</key>
```
cgwyllie, line 5: I'm not sure these keys should include the `Key` suffix?

barakbenhur1: fixed

plugins/plugin.xml
```diff
@@ -12,6 +12,7 @@
             </feature>
         </config-file>
         <source-file src="src/ios/OptimoveSDKPlugin.swift" />
+        <header-file src="src/ios/Bridging-Header-Header.h" type="BridgingHeader" />
```
cgwyllie, line 15: Think we've doubled up `-Header-Header` :)

plugins/src/ios/InitializePluginManager.m
```diff
@@ -0,0 +1,27 @@
+//
+//  InitializePluginManager.m
+//  Example_App
+//
+//  Created by Barak Ben Hur on 03/08/2022.
+//
+
+#import "InitializePluginManager.h"
+#import "Example_App-Swift.h"
```
cgwyllie, line 9: Don't think we want to import this header

plugins/src/ios/InitializePluginManager.m
```diff
@@ -0,0 +1,27 @@
...
+    dispatch_once(&onceToken, ^{
+        [NSNotificationCenter.defaultCenter addObserver: self
+                                               selector: @selector(didFinishLaunching:)
+                                                   name: UIApplicationDidFinishLaunchingNotification
+                                                 object: nil];
+    });
+}
+
++ (void) didFinishLaunching: (NSNotification*) n {
+    [[Optimove_Cordova alloc] pluginInitialize];
```
cgwyllie, line 24: We discussed not creating a second instance of this plugin, has been refactored in DDL branch so will be reconsidered as part of that scope.

cgwyllie: After further discussion, we will refactor this just now on this branch, to:

- Call a static method on Optimove_Cordova that will perform SDK initialisation
- Capture a static reference to the plugin instance from the `pluginInitialize` override in the plugin class

plugins/plugin.xml
barakbenhur1: fixed

plugins/src/ios/InitializePluginManager.m
barakbenhur1: #import "Example_App-Swift.h" removed

plugins/src/ios/InitializePluginManager.m
barakbenhur1: fixed

plugins/scripts/inject-config.js
```diff
@@ -83,8 +83,42 @@ module.exports = function injectOptimoveConfig(context) {
...
+        console.error(
+            'optimove: optimove.plist resource not found, aborting setup'
+        );
+        return;
+    }
+
+    const config = renderTemplate('optimove.xml', {
+        OPTIMOVE_CREDENTIALS: optimoveConifg.optimoveCredentials,
+        OPTIMOVE_MOBILE_CREDENTIALS: optimoveConifg.optimoveMobileCredentials
+    });
```
cgwyllie, line 117: - I think this should be using the `optimove.plist` template file for rendering
- keys of the dictionary should match the {{tags}} inside the template (e.g. template maybe can be updated to have `{{OPTIMOVE_MOBILE_CREDENTIALS}}` instead of `{{optimoveMobileCredentials}}`)

plugins/src/ios/InitializePluginManager.m
cgwyllie: OK I think we lost the part that calls the initialise of the SDK from applicationDidFinishLaunching?

plugins/scripts/inject-config.js
barakbenhur1: do you mean something like that:

    const config = renderTemplate('optimove.xml', {
        OPTIMOVE_CREDENTIALS: optimoveConifg.OPTIMOVE_CREDENTIALS,
        OPTIMOVE_MOBILE_CREDENTIALS: optimoveConifg.OPTIMOVE_MOBILE_CREDENTIALS
    });

plugins/src/ios/InitializePluginManager.m
barakbenhur1: add them back

plugins/scripts/inject-config.js
cgwyllie: I think probably more like

```
const config = renderTemplate('optimove.xml', {
    OPTIMOVE_CREDENTIALS: optimoveConifg.optimoveCredentials,
    OPTIMOVE_MOBILE_CREDENTIALS: optimoveConifg.optimoveMobileCredentials
});
```

But the plist will need updated so that the key of the dictionary here (`OPTIMOVE_CREDENTIALS`) matches the 'placeholder': `{{OPTIMOVE_CREDENTIALS}}`

barakbenhur1: done

plugins/src/ios/InitializePluginManager.m
cgwyllie: This this is back :)

plugins/src/ios/InitializePluginManager.m
```diff
@@ -0,0 +1,28 @@
...
+        [NSNotificationCenter.defaultCenter addObserver: self
+                                               selector: @selector(didFinishLaunching:)
+                                                   name: UIApplicationDidFinishLaunchingNotification
+                                                 object: nil];
+    });
+}
+
++ (void) didFinishLaunching: (NSNotification*) n {
+    Optimove_Cordova *plugin = [[Optimove_Cordova alloc] init];
+    [plugin didFinishLaunching:n];
```
cgwyllie, line 25: Is there a way we can make this method on Optimove_Cordova static so we avoid creating two Cordova plugin instances?

plugins/src/ios/InitializePluginManager.m
barakbenhur1: sorry copied it from a backup :)

barakbenhur1: fixed

plugins/src/ios/InitializePluginManager.m
barakbenhur1: done

plugins/plugin.xml
```diff
@@ -12,13 +12,14 @@
             </feature>
         </config-file>
         <source-file src="src/ios/OptimoveSDKPlugin.swift" />
+        <header-file src="src/ios/Bridging-Header.h" type="BridgingHeader" />
```
cgwyllie, line 15: We should update to add the InitializePluginManager source files (m & h)

barakbenhur1: updated

cgwyllie: OK, should the .h file be a `<header-file>` entry?

barakbenhur1: done

cgwyllie (approved):

PR #12, opened 2022-07-28
Cordova public API basic functionality android

TomBarzilay1:
The methods that was added in this PR:
public void registerUser(String userId, String email)
public void setUserEmail(String email)
public void setUserId(String userId)
public String getVisitorId()
public void reportEvent(String name)
public void reportEvent(String name, Map<String, Object> parameters)
public void reportScreenVisit(@NonNull String screenName)
public void reportScreenVisit(@NonNull String screenName, @Nullable String screenCategory)
public String getCurrentUserIdentifier()


plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -9,8 +11,167 @@
...
+                this.setUserId(userId, callbackContext);
+                return true;
+            case SET_USER_EMAIL:
+                String email = args.getString(0);
+                this.setUserEmail(email, callbackContext);
+                return true;
+            case REPORT_EVENT:
+                String eventName = args.getString(0);
+                String params = args.getString(1);
+                if (params == "null") {
```
cgwyllie, line 36: I think it should be possible to pass through an actual `null` or dictionary from JS.

See: https://github.com/Kumulos/KumulosSdkCordova/blob/81a605e4f98afe103dd6ab1e9b03b75df1aa7007/src/android/KumulosSDKPlugin.java#L201

(i.e. don't need to JSON encode it on the sender side)

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -9,8 +11,167 @@
...
+                    this.reportEvent(eventName, callbackContext);
+                } else {
+                    JSONObject jsonParams = new JSONObject(params);
+                    reportEvent(eventName, jsonParams, callbackContext);
+                }
+                return true;
+
+            case REPORT_SCREEN_VISIT:
+                String screenName = args.getString(0);
+                String screenCategory = args.getString(1);
```
cgwyllie, line 46: +1 on sending an actual `null` value, perhaps `optString()` or similar.

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -11,4 +11,4 @@
 
     let config = OptimoveConfigBuilder(optimoveCredentials: args?[0] as? String, optimoveCredentials: args?[1] as? String).enableInAppMessaging(inAppConsentStrategy: inAppConsentStrategy).build()
     Optimove.initialize(with: config)
-  }
\ No newline at end of file
+  }
```
cgwyllie, line 14: Can we revert changes to the swift side?

plugins/www/Optimove.js
```diff
@@ -6,3 +6,29 @@ var exec = require('cordova/exec');
...
     exec(success, error, 'OptimoveSDKPlugin', 'initialize', [inAppConsentStrategy]);
 };
+window.setUserId = function (userId, success, error) {
+    exec(success, error, 'OptimoveSDKPlugin', 'setUserId', [userId]);
+}
+window.setUserEmail = function (userEmail, success, error) {
+    exec(success, error, 'OptimoveSDKPlugin', 'setUserEmail', [userEmail]);
+};
+
+/* eventParams is nullable*/
```
cgwyllie, line 16: I think eventually we should try and either use TS for the plugin sources, or add a TS interfaces definition file to the project, so perhaps these annotations are redundant.

Alternately, we could use JSDoc to fully document the methods.

What do you think?

plugins/www/Optimove.js
```diff
@@ -6,3 +6,29 @@ var exec = require('cordova/exec');
...
+window.reportEvent = function (eventName, eventParams, success, error) {
+    exec(success, error, 'OptimoveSDKPlugin', 'reportEvent', [eventName, eventParams]);
+}
+
+/* screenCategory parameter is nullable*/
+window.reportScreenVisit = function (screenName, screenCategory, success, error) {
+    exec(success, error, 'OptimoveSDKPlugin', 'reportScreenVisit', [screenName, screenCategory]);
+}
+window.registerUser = function (userId, userEmail, success, error) {
```
cgwyllie, line 25: Can we use a Promise style API rather than taking success/error callbacks as method parameters?

plugins/www/Optimove.js
```diff
@@ -6,3 +6,29 @@ var exec = require('cordova/exec');
 window.initialize = function (inAppConsentStrategy, success, error) {
     exec(success, error, 'OptimoveSDKPlugin', 'initialize', [inAppConsentStrategy]);
 };
+window.setUserId = function (userId, success, error) {
```
cgwyllie, line 9: I think these should be exported from the module, rather than globally attached to the window.

The end user in the mobile app should really be consuming methods on the Optimove SDK, like `Optimove.setUserId(...)`.

Probably something like

```
module.exports = {
  setUserId: function setUseId() {},
  ...

};
```

For the public API, and then the `plugin.xml` sets the exported module name with the `clobbers` property on the JS module file.

cgwyllie: (see https://github.com/Kumulos/KumulosSdkCordova/blob/81a605e4f98afe103dd6ab1e9b03b75df1aa7007/plugin.xml)

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -9,8 +11,167 @@
...
+    private static final String REPORT_SCREEN_VISIT = "reportScreenVisit";
+    private static final String REGISTER_USER = "registerUser";
+    private static final String GET_VISITOR_ID = "getVisitorId";
+    private static final String GET_CURRENT_USER_IDENTIFIER = "getCurrentUserIdentifier";
+
     @Override
     public boolean execute(String action, JSONArray args, CallbackContext callbackContext) throws JSONException {
+        switch (action) {
+            case SET_USER_ID:
+                String userId = args.getString(0);
```
cgwyllie, line 26: General style question on these: is it better to do the logic of unpacking arguments (and in some others, branching), in the switch case, or better to pass through args to the helper method (more similar to Kumulos plugin)?

plugins/src/android/android.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
TomBarzilay1: it is auto generated, do you think we should add it to git ignore?

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -1,16 +1,205 @@
...
+        callbackContext.success();
+
+    }
+
+    private void reportScreenVisit(JSONArray args, CallbackContext callbackContext) {
+        try {
+            String screenName = args.getString(0);
+            String screenCategory = args.optString(1);
+            if (screenCategory.equals("null")) {
+                reportScreenVisit(screenName, callbackContext);
```
cgwyllie, line 138: Do we need to have these other helpers to delegate to, seems like we can replace with Optimove.report... calls directly in here to simplify?

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -1,16 +1,205 @@
 package com.optimove.android.cordova;
 
 import org.apache.cordova.CordovaPlugin;
+
+import java.nio.file.WatchEvent;
```
cgwyllie, line 5: Is this import used? Maybe give the file a format & optimise imports in Android Studio.

plugins/www/Optimove.js
```diff
@@ -3,6 +3,42 @@
...
+            exec(resolve, reject, 'OptimoveSDKPlugin', 'registerUser', [userId, userEmail])
+        });
+    },
+
+    getVisitorId: function () {
+        return new Promise((resolve, reject) => { exec(resolve, reject, 'OptimoveSDKPlugin', 'getVisitorId', []); });
+
+    },
+
+    getCurrentUserIdentifier: function () {
```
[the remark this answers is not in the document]
TomBarzilay1: this method is exposed in flutter, I checked the flutter SDK API before I started working :)

plugins/src/android/android.iml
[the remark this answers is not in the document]
cgwyllie: I'm inclined towards yes, if it can be auto-generated again by importing the module, perhaps we don't need to check the project in?

Not sure, guess it's whatever we think will be best for future development & maintenance.

plugins/www/Optimove.js
cgwyllie: OK after discussion, we will defer adding docblocks etc. until we look at using TS for the JS bridge layer.

plugins/src/android/android.iml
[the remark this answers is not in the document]
cgwyllie: (+ .idea folder)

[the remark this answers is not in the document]
cgwyllie: Hey @TomBarzilay1 think there's still some files from the .idea folder checked in, can you `git rm` them?

plugins/www/Optimove.js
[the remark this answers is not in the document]
cgwyllie: I'm thinking about naming, I know we expose it as `getCurrentUserIdentifier` in the base SDK, but perhaps `getUserId()` is closer to an analog of `setUserId()`.

What do you think?

(Happy to merge this without changing this at this point, renaming it would be minor)

@optimove-andy @k-antipochkin

plugins/src/android/android.iml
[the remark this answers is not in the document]
cgwyllie: Ah sorry, nvm!

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -1,16 +1,146 @@
...
+            callbackContext.error(e.getMessage());
+            e.printStackTrace();
+        }
+    }
+
+    private void reportScreenVisit(JSONArray args, CallbackContext callbackContext) {
+        try {
+            String screenName = args.getString(0);
+            String screenCategory = args.optString(1);
+            if (screenCategory.equals("null")) {
```
cgwyllie, line 105: Did we investigate use of args.isNull() for this?

cgwyllie (approved):

PR #13, opened 2022-08-01
Cordova swift branch rest public api functions

barakbenhur1:
add rest of public api functions


plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -30,22 +29,77 @@ import NotificationCenter
...
+    func isAvailable(command: CDVInvokedUrlCommand) {
+        self.commandDelegate.run {
+            let pluginResult = CDVPluginResult(status: .ok, messageAs: OptimoveInApp.getInboxItems())
+            self.commandDelegate.send(pluginResult, callbackId: command.callbackId)
+        }
+    }
+    
+    @objc(updateConsent:)
+    func updateConsent(command: CDVInvokedUrlCommand) {
+        OptimoveInApp.updateConsent(forUser: command.arguments[0] as? Bool ?? false)
```
cgwyllie, line 103: Can we remove these and cover it under the scope for messaging?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,105 @@ import NotificationCenter
...
+            name = n
+        }
+        
+        if let param = command.arguments[1] as? [String: Any] {
+            parameters = param
+        }
+        
+        Optimove.shared.reportEvent(name: name, parameters: parameters)
+        
+        self.commandDelegate.run {
```
cgwyllie, line 53: In all of these methods I'm not sure if we need to run this on the delegate or not. The basic plugin example doesn't post the callbacks back to the delegate queue.

Is there a particular reason to post these back, or do you think we can safely remove the `self.commandDelegate.run` calls?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,105 @@ import NotificationCenter
...
+        }
+        
+        Optimove.shared.setUserEmail(email: email)
+        
+        self.commandDelegate.run {
+            let pluginResult = CDVPluginResult(status: .ok)
+            self.commandDelegate.send(pluginResult, callbackId: command.callbackId)
+        }
+    }
 }
```
cgwyllie, line 138: Are we missing an implementation for getCurrentUserIdentifier ?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,105 @@ import NotificationCenter
         guard let config = OptimoveSDKPlugin.config else { return }
         Optimove.initialize(with: config)
     }
+    
+    @objc(reportEvent:)
+    func reportEvent(command: CDVInvokedUrlCommand) {
+        var name = ""
+        var parameters: [String: Any] = [:]
```
cgwyllie, line 41: Should the default value for this be `nil`? (I think it's an optional arg to reportEvent)

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,105 @@ import NotificationCenter
...
+        
+        if let n = command.arguments[0] as? String {
```
cgwyllie, line 43: Name is required, I think this should cause an error result from the plugin.

Can you compare with the Android implementation for reference?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,105 @@ import NotificationCenter
...
+        self.commandDelegate.run {
+            let pluginResult = CDVPluginResult(status: .ok)
+            self.commandDelegate.send(pluginResult, callbackId: command.callbackId)
+        }
+    }
+    
+    @objc(reportScreenVisit:)
+    func reportScreenVisit(command: CDVInvokedUrlCommand) {
+        guard let screenTitle = command.arguments.first as? String else {
+            let pluginResult = CDVPluginResult(status: .error, messageAs: "screenTitle is invalid")
```
cgwyllie, line 62: In general, can we just check the implementation of validation and error handling/error messages against the android implementation so they are consistent for each method?

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: Ah cool (y)

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: OK perhaps you can sync up with Tom about this to go over it with him

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: Yes, you're right.

I think it's probably better that if we guard it we can handle the validation in a better way than the Kumulos SDK (Perhaps there were/are some validations at the JS level, but not sure about this).

Let's see how it's implemented on the Optimove Cordova Android side and align with that.

plugins/src/ios/OptimoveSDKPlugin.swift
cgwyllie: OK, can you check this with Tom.

If that's true, I'm not sure we should expose it in JS at all.

The plugin will (ideally) have a unified interface in JS for both platforms.

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: will do

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: add guard to name

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: sure

barakbenhur1: will be handled in js

barakbenhur1: removed from js file

plugins/src/ios/OptimoveSDKPlugin.swift
barakbenhur1: validation and error handling/error messages will be handled in Optimove.js file

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -34,4 +34,94 @@ import NotificationCenter
         guard let config = OptimoveSDKPlugin.config else { return }
         Optimove.initialize(with: config)
     }
+    
+    @objc(reportEvent:)
+    func reportEvent(command: CDVInvokedUrlCommand) {
+        guard let name = command.arguments[0] as? String else {
+            let pluginResult = CDVPluginResult(status: .error, messageAs: "name is not invalid")
```
cgwyllie, line 41: Minor sp: "name is not valid" or "name is invalid" :)

cgwyllie: (see this copy/pasted in other places)

BBHOptimove: fixed

cgwyllie (approved):

PR #14, opened 2022-08-02
Cordova messaging android

TomBarzilay1:
added PushReceiver class, all the inApp messaging methods and functionality, and push notifications functionality


Example_App/Example - optimove.json
```diff
@@ -1,4 +1,5 @@
 {
     "optimoveCredentials": "YOUR_OPTIMOVE_CREDENTIALS",
-    "optimoveMobileCredentials": "YOUR_OPTIMOVE_MOBILE_CREDENTIALS"
+    "optimoveMobileCredentials": "YOUR_OPTIMOVE_MOBILE_CREDENTIALS",
+    "optimoveInAppConsentStrategy" : "YOUR_IN_APP_CONSENT_STRATEGY"
```
cgwyllie, line 4: I think for naming we can just have a key like `in-app-consent-strategy` or something.

Can we compare with Flutter / align at some point? (maybe alignment can wait to change iOS + Android together)

cgwyllie: We could also perhaps set the value to `auto-enroll` by default for the example.

Hm, are we using `enrol` or `enroll` spellings...

plugins/plugin.xml
```diff
@@ -26,14 +26,26 @@
...
+                <intent-filter>
+                    <action android:name="com.google.firebase.MESSAGING_EVENT" />
+                </intent-filter>
+            </service>
+            <!-- Optimove push receiver  -->
+            <receiver android:name="com.optimove.android.cordova.PushReceiver" android:exported="false">
+                <intent-filter>
+                    <action android:name="com.optimove.push.RECEIVED" />
+                    <action android:name="com.optimove.push.OPENED" />
+                    <action android:name="com.optimove.push.BUTTON_CLICKED" />
```
cgwyllie, line 44: There might be a missing type for dismissed, perhaps compare to base SDK now, not just Cordova SDK.

There were some changes that were merged into the original Kumulos Android SDK but not all x-platform ones were update.

plugins/scripts/inject-config.js
```diff
@@ -2,6 +2,24 @@
 const fs = require('fs');
 const path = require('path');
 
+function createJsonWithDefaultValues(optimoveCredentialsParam, optimoveMobileCredentialsParam ,optimoveInAppConsentStrategy) {
+    var optimoveCredentials ="";
+    var optimoveMobileCredentials = "";
+    
+    if (optimoveCredentialsParam) {
```
cgwyllie, line 9: Worth using isEmpty here? What values can the args have?

cgwyllie: Perhaps we can use some tertiary condition when building the returned object, for example

```
return {
  OPTIMOVE_CREDS: isEmpty(optimoveCredentials) ? "" : optimoveCredentials,
...}
```

What do you think?

(i.e. can rename function params to drop `Param` suffix and lose the local vars)

Example_App/Example - optimove.json
cgwyllie: (actually maybe disabled is better)

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -10,12 +10,17 @@
...
 import androidx.annotation.Nullable;
+
 import com.optimove.android.Optimove;
 import com.optimove.android.OptimoveConfig;
+import com.optimove.android.optimobile.OptimoveInApp;
 
 public class OptimoveInitProvider extends ContentProvider {
     private static final String OPTIMOVE_CREDENTIALS = "optimoveCredentials";
     private static final String OPTIMOVE_MOBILE_CREDENTIALS = "optimoveMobileCredentials";
+    private static final String IN_APP_AUTO_ENROLL = "autoEnroll";
```
cgwyllie, line 21: I think these don't match the values in JSON -> XML ?

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -74,4 +92,12 @@ private String getStringConfigValue(String packageName, Resources resources, Str
 
         return resources.getString(resId);
     }
+
+    @Nullable
+    private String handleNullValues(String credentials) {
+        if (credentials.equals("undefined") || credentials.equals("null") || credentials.isEmpty()) {
```
cgwyllie, line 98: Can we refactor so the file writes empty string values then perhaps handle with TextUtils.isNullOrEmpty() ?

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -48,11 +86,142 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
...
+                if (result == OptimoveInApp.InboxMessagePresentationResult.PRESENTED) {
+                    callbackContext.success();
+                    return;
+                } else {
+                    break;
+                }
+            }
+        }
+
+        callbackContext.error("Message not found or not available");
```
cgwyllie, line 176: Do we want to consider passing back the in app presentation result value here?

cgwyllie: (if available)

Example_App/Example - optimove.json
TomBarzilay1: sure I'll look at the flutter sdk

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: I changed it to auto-enroll

plugins/plugin.xml
TomBarzilay1: I added dismissed action :)

plugins/scripts/inject-config.js
TomBarzilay1: I actually think it's better that way because it covers more cases, for example the case which the args are undefined, and it also covers the case that the string is empty or null, what do you think? (the args will behave according to the optimove.json, if we put a key without value it will be null or empty and if don't put a key it's undefined)

Example_App/Example - optimove.json
```diff
@@ -1,4 +1,5 @@
 {
     "optimoveCredentials": "YOUR_OPTIMOVE_CREDENTIALS",
-    "optimoveMobileCredentials": "YOUR_OPTIMOVE_MOBILE_CREDENTIALS"
+    "optimoveMobileCredentials": "YOUR_OPTIMOVE_MOBILE_CREDENTIALS",
+    "inAppConsentStrategy" : "YOUR_IN_APP_CONSENT_STRATEGY"
```
cgwyllie, line 4: In the example file, should we put the 'disabled' value?

TomBarzilay1: I'm sorry I'm not sure I understand the question, where should we put disabled?

plugins/scripts/inject-config.js
cgwyllie: I think isEmpty should handle null/empty string/undefined values too.

I guess truthiness works with some coercion in JS, so it's roughly equivalent. I'm maybe thinking how to simplify the structure of variables + parameters, but maybe it's not a big deal.

One further consideration might be testing the values is a string before passing it through. What do you think?

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -10,12 +10,17 @@
...
 import com.optimove.android.Optimove;
 import com.optimove.android.OptimoveConfig;
+import com.optimove.android.optimobile.OptimoveInApp;
 
 public class OptimoveInitProvider extends ContentProvider {
     private static final String OPTIMOVE_CREDENTIALS = "optimoveCredentials";
     private static final String OPTIMOVE_MOBILE_CREDENTIALS = "optimoveMobileCredentials";
+    private static final String IN_APP_AUTO_ENROLL = "auto-enroll";
+    private static final String IN_APP_EXPLICIT_BY_USER = "explicit-by-user";
+    private static final String KEY_IN_APP_CONSENT_STRATEGY = "inAppConsentStrategy";
```
cgwyllie, line 23: Structure: can we group the keys together / common prefixes if they're all keys? (i.e. OPTIMOVE_CREDENTIALS -> KEY_...)

TomBarzilay1: sure thing

plugins/www/Optimove.js
```diff
@@ -1,44 +1,188 @@
...
+  if (typeof currentConfig[handlerName] === "function") {
+    currentConfig[handlerName](message.data);
+  } else {
+    console.log(`Optimove: No handler defined for '${message.type}' event`);
+  }
+}
 
 const Optimove = {
-    setUserId: function (userId) {
+  initBaseSdk: function () {
```
cgwyllie, line 36: I'm not sure this is exposed on the public JS interface, I think it only needs called when the developer sets their callbacks for push/in-app etc.

plugins/scripts/inject-config.js
TomBarzilay1: sure sounds like an important thing to check., in the situation where inside the optimove.json there won't be a key for the property (for ex inAppConsentStrategy), the value will be undefined, should I cover undefined values also or we can just assume that the json will be in a right format?

plugins/www/Optimove.js
TomBarzilay1: so should I create a js object that will represent the configuration of the optimove sdk, and add the handlers and this function also to that object?

Example_App/Example - optimove.json
TomBarzilay1: do you mean to put 'disabled' value to all the properties that should be missing?

plugins/www/Optimove.js
vvoicehovics: havent looked much at what does what, but is this just renaming? Base sdk is our OptimoveAndroid, so, `initBaseSdk` shall be an internal method. But could have `initBaseSdk` -> `initialize`?

vvoicehovics: havent looked much at what does what, but is this just renaming? Base sdk is our OptimoveAndroid, so, `initBaseSdk` shall be an internal method. Could have `initBaseSdk` -> `initialize`?

Do we cuttently set push/in-app handlers?

Example_App/www/js/index.js
```diff
@@ -17,101 +17,266 @@
...
 }
 
 function error(errorMessage) {
-    console.error(errorMessage);
+  console.error(errorMessage);
 }
 
+document
+  .getElementById("init-base-sdk-button")
+  .addEventListener("click", initBaseSdk);
```
vvoicehovics, line 47: does it just happen when app starts, why need button for it?

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: sure, done :)

plugins/src/android/OptimoveSDKPlugin.java
TomBarzilay1: I think in the discussion yesterday we decided to keep it the way it is

Example_App/Example - optimove.json
TomBarzilay1: I changed "YOUR_IN_APP_CONSENT_STRATEGY" to in-app-disabled

plugins/www/Optimove.js
```diff
@@ -1,44 +1,204 @@
+var exec = require("cordova/exec");
 
+var inAppInboxUpdatedHandler;
 
-var exec = require('cordova/exec');
+var currentConfig = {
+  pushReceivedHandler: pushReceivedHandler, //function that receives one argument, a push message object
```
cgwyllie, line 6: What are these implementations currently set to, I don't see `pushReceivedHandler` etc. definitions.

Perhaps we declare some 'no-op' function `function noop() {}` and set a reference to it for all of these by default.

What do you think?

plugins/www/Optimove.js
```diff
@@ -1,44 +1,204 @@
...
+  });
+}
+
+function nativeMessageHandler(message) {
+  if (!message || typeof message === "string") {
+    return;
+  }
+  
+  const handlerName = `${message.type}Handler`;
+  alert(handlerName);
```
cgwyllie, line 42: Can we chop alerting from the plugin implementation?

plugins/www/Optimove.js
cgwyllie: (alternately setting them explicitly to undefined or null is fine)

plugins/www/Optimove.js
```diff
@@ -1,44 +1,204 @@
...
+    console.log(`Optimove: No handler defined for '${message.type}' event`);
+  }
+}
+
+function success(message = "success!") {
+  console.log(message);
+}
+function error(message = "error") {
+  console.log(message);
+}
```
cgwyllie, line 64: Are these used?

plugins/www/Optimove.js
```diff
@@ -1,44 +1,204 @@
...
+    });
+  },
+
+  setOnInboxUpdatedHandler: function (handler) {
+      inAppInboxUpdatedHandler = handler;
+  },
+  
+  setPushOpenedHandler(pushOpenedHandler) {
+    currentConfig["pushOpenedHandler"] = pushOpenedHandler;
+    checkIfPendingPushExists().then(success,error);
```
cgwyllie, line 193: Not sure if this needs to chain a success/error callback?

plugins/www/Optimove.js
TomBarzilay1: I think the noop is a good option, I'll add it now

plugins/www/Optimove.js
TomBarzilay1: yeah sure, sorry I forgot to delete some debug prints

plugins/www/Optimove.js
TomBarzilay1: hmm although it's an async function?

plugins/www/Optimove.js
TomBarzilay1: maybe we could give a different no op implementation for each function depending on its parameters? or it's good enough one no op function for all?

plugins/scripts/inject-config.js
```diff
@@ -2,6 +2,27 @@
 const fs = require('fs');
 const path = require('path');
 
+function createJsonWithDefaultValues(optimoveCredentials, optimoveMobileCredentials , inAppConsentStrategy) {
```
cgwyllie, line 5: I was thinking more about this method, and I'm not actually sure we will ever need this because configuration is actually validated earlier in the process.

I think there's the `isValidConfig` function that won't let the resources be generated if the JSON wasn't valid.

So, do we need to keep this or is it better to refactor it out?

TomBarzilay1: hmm I think I added this method to handle null or undefined credentials, which are legal if just one them is undefined, this method just "making sure" that the Java side will get an empty string instead of null or undefined

TomBarzilay1: so I think it's necessary but maybe there is a better way of doing it, what do you think?

plugins/www/Optimove.js
TomBarzilay1: nope, I deleted them

plugins/www/Optimove.js
TomBarzilay1: I removed the success and error

plugins/scripts/inject-config.js
```diff
@@ -2,6 +2,27 @@
...
+    if (!isString(inAppConsentStrategy) || isEmpty(inAppConsentStrategy)) {
```
vvoicehovics, line 6: is this check necessary? when we validate config we already make sure in-app strategy is good

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -10,30 +10,49 @@
...
     @Override
     public boolean onCreate() {
         Application app = (Application) getContext().getApplicationContext();
         String packageName = app.getPackageName();
         Resources resources = app.getResources();
-        String optimoveCredentials = getStringConfigValue(packageName, resources, OPTIMOVE_CREDENTIALS);
-        String optimoveMobileCredentials = getStringConfigValue(packageName, resources, OPTIMOVE_MOBILE_CREDENTIALS);
-        if (TextUtils.isEmpty(optimoveCredentials) || TextUtils.isEmpty(optimoveMobileCredentials)) {
-            return true;
+        String optimoveCredentials = getStringConfigValue(packageName, resources,KEY_OPTIMOVE_CREDENTIALS);
```
vvoicehovics, line 31: do we want formatting js/java? Maybe worth having a separate PR

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -74,4 +93,9 @@ private String getStringConfigValue(String packageName, Resources resources, Str
 
         return resources.getString(resId);
     }
+
+    @Nullable
+    private String handleNullValues(String credentials) {
```
vvoicehovics, line 98: minor random stuff. maybe a little neater could combining `handleNullValues` and `getStringConfigValue` to get 

```
private String readConfigValue(String packageName, Resources resources, String key) {
 int resId = resources.getIdentifier(key, "string", packageName);
 if (0 == resId) {
    return null;
 }

  String value = resources.getString(resId);
  return TextUtils.isEmpty(value) ? null : value;
}
```

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -74,4 +93,9 @@ private String getStringConfigValue(String packageName, Resources resources, Str
 
         return resources.getString(resId);
     }
+
+    @Nullable
```
vvoicehovics, line 97: it's a style thing probably, but this annotation applies to the return type, so, mb nicer to put it before that , mm, for consistency? I think in android we have it that way

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -10,30 +10,49 @@
...
-        String optimoveCredentials = getStringConfigValue(packageName, resources, OPTIMOVE_CREDENTIALS);
-        String optimoveMobileCredentials = getStringConfigValue(packageName, resources, OPTIMOVE_MOBILE_CREDENTIALS);
-        if (TextUtils.isEmpty(optimoveCredentials) || TextUtils.isEmpty(optimoveMobileCredentials)) {
-            return true;
+        String optimoveCredentials = getStringConfigValue(packageName, resources,KEY_OPTIMOVE_CREDENTIALS);
+        String optimoveMobileCredentials = getStringConfigValue(packageName, resources,KEY_OPTIMOVE_MOBILE_CREDENTIALS);
+        optimoveCredentials = handleNullValues(optimoveCredentials);
+        optimoveMobileCredentials = handleNullValues(optimoveMobileCredentials);
+        String inAppConsentStrategy = getStringConfigValue(packageName, resources, KEY_IN_APP_CONSENT_STRATEGY);
+        assert (optimoveCredentials != null || optimoveMobileCredentials != null);
```
vvoicehovics, line 36: shall we use `assert` in prod code?

https://softwareengineering.stackexchange.com/questions/137158/is-it-better-to-use-assert-or-illegalargumentexception-for-required-method-param

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -138,4 +320,142 @@ private void getCurrentUserIdentifier(CallbackContext callbackContext) {
             callbackContext.error("current user identifier is null");
         }
     }
+
+    private void setHandlersCallBackContext(CallbackContext callbackContext) {
+        jsCallbackContext = callbackContext;
+        PluginResult result = new PluginResult(PluginResult.Status.OK);
+        result.setKeepCallback(true);
```
vvoicehovics, line 327: just a question, what does this do?

TomBarzilay1: this function save a reference to a callback context in which through it we send messages back to the js (like receiving an event in the java code and send it to a handler in the js code)

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: ok I'll change it

plugins/scripts/inject-config.js
TomBarzilay1: yes I agree, I thought to double check here, do you think we should get rid of it?

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -138,4 +320,142 @@ private void getCurrentUserIdentifier(CallbackContext callbackContext) {
...
+        }
+
+        callbackContext.success(results);
+    }
+
+    private void inAppMarkAllInboxItemsAsRead(CallbackContext callbackContext) {
+        try {
+            OptimoveInApp.getInstance().markAllInboxItemsAsRead();
+            callbackContext.success();
+        } catch (Exception e) {
```
vvoicehovics, line 441: implementation in kumulos SDK is

```
    private void inAppMarkAllInboxItemsRead(CallbackContext callbackContext) {
        boolean result = KumulosInApp.markAllInboxItemsAsRead(cordova.getContext());
        if (result) {
            callbackContext.success();
        }
        else {
            callbackContext.error("Failed to mark all messages as read");
        }
    }
```

Can you check, do we ever throw exception here? Was kumulos implementation wrong, what's the story?

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: done, I agree it is neater that way

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -48,11 +87,154 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
...
+            }
+        });
+    }
+
+    private void inAppMarkAsRead(JSONArray args, CallbackContext callbackContext) {
+        try {
+            JSONObject inboxItem = args.optJSONObject(0);
+            int messageId = -1;
+            if (inboxItem != null) {
+                messageId = args.optInt(0);
```
vvoicehovics, line 213: here i dont get these conditionals a bit. We take 1st argument json object. If it is not null, we take 1st argument int?

is intention allow passing either id and inboxItem?

TomBarzilay1: it should be " messageId = inboxItem.optInt("id");"
I fixed it now

vvoicehovics: updated to pass id into javaland

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -138,4 +275,155 @@ private void getCurrentUserIdentifier(CallbackContext callbackContext) {
...
+    private void setHandlersCallBackContext(CallbackContext callbackContext) {
+        jsCallbackContext = callbackContext;
+        PluginResult result = new PluginResult(PluginResult.Status.OK);
+        result.setKeepCallback(true);
+        callbackContext.sendPluginResult(result);
+    }
+
+    private void checkIfPendingPushExists() {
+        if (null != pendingPush) {
+            OptimoveSDKPlugin.sendMessageToJs("pushOpened",
```
vvoicehovics, line 288: to trigger pushOpenedHandler, here we need jsCallbackContext set. Is there a possibility of race/sequencing issues? 

Now we add 2 listeners for 'deviceready' event: 
1) one in Optimove.js calls `init`, which sets `jsCallbackContext`
2) and one in index.js sets `pushOpenedHandler` which checks for pendingPush. 

Is there a guarantee that 1) runs before 2) ?

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: to stay consistent with `inAppMarkAsRead` send error response when result is false

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: i was asking more about this `result.setKeepCallback(true);`

TomBarzilay1: this function keeps the contextCallback "alive" so we can use it multiple times

TomBarzilay1: https://ourcodeworld.com/articles/read/33/how-to-execute-a-cordova-callback-more-than-once-in-java-android-

vvoicehovics: 👍

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: updated to make it more robust racing-wise

plugins/src/android/PushReceiver.java
```diff
@@ -0,0 +1,128 @@
...
+
+        return message;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    protected void onPushOpened(Context context, PushMessage pushMessage) {
+        try {
+            Optimove.getInstance().pushTrackOpen(pushMessage.getId());
+        } catch (Exception e) {
```
vvoicehovics, line 53: this is different from base android / flutter. There we catch `Optimobile.UninitializedException` . Better to use more specific type?

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: Ok, does `OptimoveSDKPlugin.jsCallbackContext.error(...` invalidate the context then? based on this it does 

https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/CallbackContext.java#L121

Example_App/www/js/index.js
```diff
@@ -21,97 +21,233 @@
...
+function inAppMarkAllInboxItemsAsRead() {
+  Optimove.inAppMarkAllInboxItemsAsRead().then(success, error);
+}
+
+document
+  .getElementById("in-app-mark-inbox-item-as-read-button")
+  .addEventListener("click", inAppMarkAsRead);
+
+function inAppMarkAsRead() {
+  Optimove.inAppMarkAsRead(
```
vvoicehovics, line 200: this accepts item, not an id. Worth updating ExampleApp?

plugins/src/android/optimove.gradle
```diff
@@ -19,7 +19,9 @@ dependencies{
     implementation("androidx.work:work-runtime-ktx:$work_version") {
         force = true
     }
-  
+    implementation platform('com.google.firebase:firebase-bom:28.2.0')
+    implementation 'com.google.firebase:firebase-messaging'
```
vvoicehovics, line 23: firebase added 2 times? also, com.optimove.android:optimove-android:5.+ shall be 6.+, not sure if is updated somewhere else

plugins/src/android/optimove.gradle
```diff
@@ -35,4 +37,18 @@ buildscript {
...
+}
+ext.postBuildExtras = {
+    if (file("./google-services.json").exists()) {
+        if (project.extensions.findByName('googleServices') == null) {
+            // apply plugin: 'com.google.gms.google-services'
+            // class must be used instead of id(string) to be able to apply plugin from non-root gradle file
+            apply plugin: com.google.gms.googleservices.GoogleServicesPlugin
+        }
+    }
+    dependencies {
```
vvoicehovics, line 49: in kumulos cordova sdk we had

```
def firebaseBomVersion = '28.2.0'
dependencies {
        if (project.hasProperty('KumulosFirebaseBoMVersion')){
            firebaseBomVersion = project.ext.get("KumulosFirebaseBoMVersion")
        }

        implementation platform("com.google.firebase:firebase-bom:$firebaseBomVersion")
        implementation 'com.google.firebase:firebase-messaging'
    }
```

the point of `KumulosFirebaseBoMVersion` was letting consumers of the SDK override bom version to their liking. Do we need smth similar @cgwyllie ?

plugins/src/android/optimove.gradle
```diff
@@ -35,4 +37,18 @@ buildscript {
         
         
     }
+}
+ext.postBuildExtras = {
```
vvoicehovics, line 41: // use postBuildExtras to make sure the plugin is applied after
// cdvPluginPostBuildExtras. Therefore if googleServices is added
// to cdvPluginPostBuildExtras somewhere else, the plugin execution
// will be skipped and project build will be successfull

i think this comment is useful

plugins/www/Optimove.js
```diff
@@ -1,39 +1,188 @@
+var exec = require("cordova/exec");
 
+var inAppInboxUpdatedHandler;
```
vvoicehovics, line 3: is there a reason why this one is special? Why it doesnt reside on config like all others? 

I think in kumulos sdk the reason was that we wanted to be able to set it to null, but could not modify after config passed into initialisation. Here handlers are not part of config

plugins/www/Optimove.js
vvoicehovics: why noop is better than null? If consumers decide to unset a handler they are likely to set to null, not to empty function. Then we need to handle it being null anyway/

for example here
```
  if (typeof currentConfig[handlerName] === "function") {
    currentConfig[handlerName](message.data);
  } else {
    console.log(`Optimove: No handler defined for '${message.type}' event`);
  }
```
with noop we will never log 'handler not defined'

plugins/src/android/PushReceiver.java
TomBarzilay1: done

Example_App/www/js/index.js
TomBarzilay1: sure, done

plugins/src/android/optimove.gradle
TomBarzilay1: updated dependencies 👍

plugins/src/android/optimove.gradle
TomBarzilay1: done :+1

plugins/src/android/OptimoveSDKPlugin.java
cgwyllie: @vvoicehovics sorry think perhaps the line is messed up, do you mean to do with the setting of the `finished` field on the context in `sendPluginResult` because it is a new result instance?

So to resolve that we should create a new PluginResult ourselves & set to keep context in any error case?

Maybe it doesn't matter, are there any non-terminal error states that could be trapped?

plugins/src/android/optimove.gradle
cgwyllie: Hm, that's a good question.

Perhaps we should replicate the style we had in Kumulos to allow overriding as you say.

We could theoretically add it later if someone has a conflict, but doesn't seem like a large change to make so perhaps good to do now.

Did we document how to set/override it, and should we add to Optimove docs when we have them?

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: I think reusing `OptimoveSDKPlugin.jsCallbackContext.error()` is not going to work because

1) yes, when we call that it does 

```
  public void error(JSONObject message) {
        sendPluginResult(new PluginResult(PluginResult.Status.ERROR, message));
    }
```
(without setKeepCallback). This would invalidate context. We could work around, but mb we just avoid using `OptimoveSDKPlugin.jsCallbackContext...`. because
2a) i dont kno wto which callback it would report error
2b) we dont really need to report errors there

soooo, propose to just only use `OptimoveSDKPlugin.jsCallbackContext` for handlers

plugins/src/android/optimove.gradle
TomBarzilay1: sure, yeah I'll add it to the docs also

TomBarzilay1: done :)

plugins/www/Optimove.js
TomBarzilay1: done

plugins/src/android/OptimoveSDKPlugin.java
TomBarzilay1: we eventually used sendPluginResult to report the error which doesn't invalidate the context

vvoicehovics (approved):

PR #15, opened 2022-08-10
Deferred deep linking android

TomBarzilay1:
Add DDL for android platform. To enable DDL users of the plugin would need to:

1. add `"enableDeferredDeepLinking": true` to optimove.json
2. add to `config.xml`
 
```
<platform name="android">
	<config-file parent="/manifest/application/activity" target="AndroidManifest.xml">
		<intent-filter android:label="deepLabel" android:autoVerify="true">
			<action android:name="android.intent.action.VIEW" />
			<category android:name="android.intent.category.DEFAULT" />
			<category android:name="android.intent.category.BROWSABLE" />
			<data android:scheme="https" android:host="cordova-example-optimove.lnk.click"/>
		</intent-filter>
	</config-file>

	<edit-config file="AndroidManifest.xml" target="/manifest/application/activity/intent-filter/category/[@android:name='android.intent.category.LAUNCHER']/../.." mode="merge">
		<activity android:name="OptimoveMainActivity"/>
	</edit-config>
</platform>
```

This way, 
1. Customers have control of DDL manifest configuration  (not only host, but more). 
2. `OptimoveMainActivity` extending `MainActivity` (that is generated by Cordova) is automatically copied and used in the manifest. Customers don't need to add `onWindowFocusChanged` or `onNewIntent` every time platform is added. If automatic setup goes wrong they can always remove `edit-config` tag and use activity of their choice.


plugins/src/android/OptimoveInitProvider.java
```diff
@@ -100,4 +111,32 @@ private String handleNullValues(String credentials) {
...
     }
+
+    private DeferredDeepLinkHandlerInterface getDDLHandlerInterface() {
+        return (context, resolution, link, data) -> {
+            try {
+                JSONObject dataJson = null;
+                if (null != data) {
+                    JSONObject deepLinkContent = null;
+                    if (data.content != null) {
+                        deepLinkContent.put("title", data.content.title);
```
vvoicehovics, line 122: null pointer exception?

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -100,4 +111,32 @@ private String handleNullValues(String credentials) {
         }
         return credentials;
     }
+
+    private DeferredDeepLinkHandlerInterface getDDLHandlerInterface() {
```
vvoicehovics, line 115: function name, it returns handler, not an interface

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -100,4 +111,32 @@ private String handleNullValues(String credentials) {
...
+        return (context, resolution, link, data) -> {
```
vvoicehovics, line 116: could have parameter types here?

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -13,14 +13,21 @@
...
+import org.json.JSONObject;
+
 public class OptimoveInitProvider extends ContentProvider {
     private static final String OPTIMOVE_CREDENTIALS = "optimoveCredentials";
     private static final String OPTIMOVE_MOBILE_CREDENTIALS = "optimoveMobileCredentials";
     private static final String IN_APP_AUTO_ENROLL = "auto-enroll";
     private static final String IN_APP_EXPLICIT_BY_USER = "explicit-by-user";
     private static final String KEY_IN_APP_CONSENT_STRATEGY = "inAppConsentStrategy";
+    private static final String ENABLE_DEFERRED_DEEP_LINKING = "enableDeferredDeepLinking";
+    private static final String TRUE = "true";
```
vvoicehovics, line 29: this is weird, necessary?

plugins/www/Optimove.js
```diff
@@ -172,6 +173,7 @@ const Optimove = {
     currentConfig["pushReceivedHandler"] = pushReceivedHandler;
     currentConfig["pushOpenedHandler"] = pushOpenedHandler;
     currentConfig["inAppDeepLinkHandler"] = inAppDeepLinkHandler;
+    currentConfig['deepLinkHandler'] = deepLinkHandler;
```
vvoicehovics, line 176: 4th arg missing?

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: fixed

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: I changed the name to getDDLHandler

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: yep, I added parameter types

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: no, I deleted it

Example_App/www/js/index.js
```diff
@@ -45,6 +45,11 @@ function error(errorMessage) {
 }
 
 
+Optimove.setOnInboxUpdatedHandler(onInboxUpdated);
```
vvoicehovics, line 48: is it necessary here? Is it not set in `onDeviceReady`?

plugins/plugin.xml
```diff
@@ -31,6 +31,14 @@
                 <param name="android-package" value="com.optimove.android.cordova.OptimoveSDKPlugin" />
             </feature>
             <preference name="AndroidXEnabled" value="true" />
+        </config-file >
+        <config-file parent="/manifest/application/activity" target="AndroidManifest.xml">
+            <intent-filter android:label="deepLabel" android:autoVerify="true">
+                 <action android:name="android.intent.action.VIEW" />
+                 <category android:name="android.intent.category.DEFAULT" />
+                 <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="https" android:host="cordova-example-optimove.lnk.click"/>
```
vvoicehovics, line 40: is this part of SDK? But host must be per customer. What's the plan for this? Could
1) ask customers to add to manifest
2) also somehow read from some config and set when plugin is added

plugins/scripts/inject-config.js
```diff
@@ -15,105 +15,110 @@ function createJsonWithDefaultValues(optimoveCredentials, optimoveMobileCredenti
             ? optimoveMobileCredentials
             : "",
         IN_APP_STRATEGY: inAppConsentStrategy,
+         ENABLE_DEFERRED_DEEP_LINKING:
+      !isEmpty(enableDeferredDeepLinking) && isString(enableDeferredDeepLinking)
+        ? enableDeferredDeepLinking
+        : ""
```
vvoicehovics, line 21: can use false here instead of "" ? In OptimoveInitProvider we then do `Boolean.parseBoolean(enableDeferredDeepLinking)`

plugins/plugin.xml
TomBarzilay1: yes you are right, maybe its best to be aligned with flutter and ask customers to add it to the manifest

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -95,4 +106,33 @@ public int update(@NonNull Uri uri, @Nullable ContentValues values, @Nullable St
...
     }
+
+    private DeferredDeepLinkHandlerInterface getDDLHandler() {
+        return (Context context, DeferredDeepLinkHelper.DeepLinkResolution resolution, String link,
+                DeferredDeepLinkHelper.DeepLink data) -> {
+            try {
+                JSONObject dataJson = null;
+                if (null != data) {
+                    JSONObject deepLinkContent = new JSONObject();
+                    if (data.content != null) {
```
vvoicehovics, line 117: if i read correct, content is not optional in `DeferredDeepLinkHelper.DeepLink`?

plugins/scripts/inject-config.js
TomBarzilay1: yep you are right, done

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -95,4 +106,33 @@ public int update(@NonNull Uri uri, @Nullable ContentValues values, @Nullable St
...
+                    dataJson.put("content", deepLinkContent);
+                    dataJson.put("url", data.url);
+
+                }
+                JSONObject deepLink = new JSONObject();
+                deepLink.put("link", link);
+                deepLink.put("resolution", resolution.ordinal());
+                deepLink.put("data", dataJson);
+                OptimoveSDKPlugin.pendingDDL = deepLink;
+                OptimoveSDKPlugin.sendMessageToJs("deepLink", deepLink);
```
vvoicehovics, line 132: a general question for Cordova. in other sdks we used to prefix message names to reduce possibility of collisions with other plugins sending java->js messages. For example, `deepLink` --> `optimove.deep-linking.deepLink`)

Any value in doing same? If yes, better to do in a separate 'finishing' PR as issue for all messages

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -95,4 +106,33 @@ public int update(@NonNull Uri uri, @Nullable ContentValues values, @Nullable St
...
+                    dataJson.put("data", data.data != null ? data.data : null);
+                    dataJson.put("content", deepLinkContent);
+                    dataJson.put("url", data.url);
+
+                }
+                JSONObject deepLink = new JSONObject();
+                deepLink.put("link", link);
+                deepLink.put("resolution", resolution.ordinal());
+                deepLink.put("data", dataJson);
+                OptimoveSDKPlugin.pendingDDL = deepLink;
```
vvoicehovics, line 131: do we always save pendingDDL _and_ send message to js? If callback context was initialised, does it mean we call handler twice?

plugins/src/android/OptimoveSDKPlugin.java
```diff
@@ -128,11 +131,25 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
...
     }
 
+    private void checkIfPendingDDLExists(CallbackContext callbackContext) {
+        if (jsCallbackContext == null) {
+            jsCallbackContext = callbackContext;
+        }
+        if (null != pendingDDL) {
+            OptimoveSDKPlugin.sendMessageToJs("deepLink", pendingDDL);
+            pendingDDL = null;
+            pendingActionId = null;
```
vvoicehovics, line 149: pendingActionId is only relevant for push, it has to do with action buttons in push notifications

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -20,7 +20,11 @@ import NotificationCenter
             return nil
         }
         
-        let config = OptimoveConfigBuilder(optimoveCredentials: configValues[optimoveCredentialsKey], optimobileCredentials: configValues[optimoveMobileCredentialsKey])
+        let optimoveCredentials = configValues[optimoveCredentialsKey]?.isEmpty == true ? nil : configValues[optimoveCredentialsKey]
+        
+        let optimobileCredentials = configValues[optimoveMobileCredentialsKey]?.isEmpty == true ? nil : configValues[optimoveMobileCredentialsKey]
```
vvoicehovics, line 25: is this some bug fix for Swift or why in android DDL pr?

plugins/www/Optimove.js
```diff
@@ -17,6 +17,7 @@ function init() {
 
 function resume() {
   setContext();
+  checkIfPendingDDLExists().then(null, (errMsg) => { console.error(errMsg); });
```
vvoicehovics, line 20: why this is checked in `resume` vs for push we check it after handler is set. Any particular reason?

in general checking for this pending stuff is a total mess. Ideally would be better to have some event sync which would deliver a message after all is ready. We do that in Optimove Flutter and Kumulos react native i think. Not proposing changing it now

plugins/plugin.xml
TomBarzilay1: I added a section to the wiki with an explanation of what should be done [https://github.com/optimove-tech/Optimove-SDK-Cordova/wiki/deferred-deep-linking](url)

plugins/src/ios/OptimoveSDKPlugin.swift
TomBarzilay1: not sure if Barak worked on my branch or not, maybe it was one of the merges from dev, I will check that

plugins/www/Optimove.js
TomBarzilay1: I changed it to be like with the pending push

plugins/src/android/OptimoveSDKPlugin.java
TomBarzilay1: I deleted it 👍

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: I changed it to be like with the pending push

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: worth a discussion to understand the value of it

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: I deleted the unnecessary checks for null values

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: worth a discussion to understand the value of it

plugins/src/android/OptimoveInitProvider.java
TomBarzilay1: I deleted the unnecessary checks for null values

vvoicehovics: `DeferredDeepLinkHelper.DeepLink data` is nullable. i meant data key inside of it. tweaked

plugins/src/android/OptimoveSDKPlugin.java
vvoicehovics: mb didnt push? deleted also ;)

plugins/src/ios/OptimoveSDKPlugin.swift
vvoicehovics: reverted, this is prob added in some other pr, mb error handling

plugins/src/android/OptimoveInitProvider.java
vvoicehovics: added to the final TODO list

plugins/plugin.xml
vvoicehovics: not sure about Flutter, but cordova has many ways to tweak manifest etc. So,

1. For the manifest DDL configuration clients shall add DDL configuration to `config.xml`. This way done only once for app and they have full control of schemes / host / ...
2. ...

Example_App/config.xml
```diff
@@ -1,11 +1,30 @@
...
+		<config-file parent="/manifest/application/activity" target="AndroidManifest.xml">
+			<intent-filter android:label="deepLabel" android:autoVerify="true">
+				<action android:name="android.intent.action.VIEW" />
+				<category android:name="android.intent.category.DEFAULT" />
+				<category android:name="android.intent.category.BROWSABLE" />
+				<data android:scheme="https" android:host="cordova-example-optimove.lnk.click"/>
+			</intent-filter>
+		</config-file>
+
+		<edit-config file="AndroidManifest.xml" target="/manifest/application/activity/intent-filter/category/[@android:name='android.intent.category.LAUNCHER']/../.." mode="merge">
```
vvoicehovics, line 26: took me a while to find this https://python.readthedocs.io/en/latest/library/xml.etree.elementtree.html

Could not filter activity by name because name changes when editing config and filtering options are limited. So, instead we pick activity with LAUNCHER intent

plugins/src/android/OptimoveInitProvider.java
```diff
@@ -23,6 +28,8 @@ public class OptimoveInitProvider extends ContentProvider {
     private static final String IN_APP_AUTO_ENROLL = "auto-enroll";
     private static final String IN_APP_EXPLICIT_BY_USER = "explicit-by-user";
 
+    private static final String ENABLE_DEFERRED_DEEP_LINKING = "enableDeferredDeepLinking";
```
cgwyllie, line 31: Just realising the resources seem to be shared, whilst theoretically unlikely to have key collisions, is it worth prefixing keys in resources?

plugins/src/android/OptimoveInitProvider.java
cgwyllie: In general a pointer on naming of messages that model events would be to think about declaring names in a past tense, i.e. `deepLinkMatched` etc., events are 'facts', things that have happened in the state of the program rather than "commands" to instruct some other module to do something :)

Not necessary to update unless you feel strongly about it, just something to think about.

plugins/src/android/optimove.gradle
```diff
@@ -14,12 +14,12 @@ android {
...
 dependencies{
     implementation 'com.optimove.android:optimove-android:6.+'
-       def work_version = "2.6.0"
-    // Force WorkManager 2.6.0 for transitive dependency
-    implementation("androidx.work:work-runtime-ktx:$work_version") {
-        force = true
+   
+    api(group: "androidx.work", name: "work-runtime") {
+    version {
+        strictly "2.7.0-alpha04"
```
cgwyllie, line 20: ?

plugins/scripts/inject-config.js
```diff
@@ -1,200 +1,255 @@
 
 const fs = require('fs');
 const path = require('path');
+const elementtree = require('elementtree');
```
cgwyllie, line 4: Is this dep available by default in Cordova installs or should we have a dependency on it?

plugins/scripts/inject-config.js
```diff
@@ -1,200 +1,255 @@
...
+        ? optimoveCredentials
+        : "",
+    OPTIMOVE_MOBILE_CREDENTIALS:
+      !isEmpty(optimoveMobileCredentials) &&
+      isString(optimoveMobileCredentials)
+        ? optimoveMobileCredentials
+        : "",
+    IN_APP_STRATEGY: inAppConsentStrategy,
+    ENABLE_DEFERRED_DEEP_LINKING:
+      enableDeferredDeepLinking === 'true' || enableDeferredDeepLinking === true
```
cgwyllie, line 149: why do we handle this as both string and boolean?

vvoicehovics: just they could type any, but guess no reason. updated to check only for bool

plugins/scripts/inject-config.js
vvoicehovics: i think that's what Cordova uses to parse plugin.xml and config.xml

plugins/src/android/optimove.gradle
vvoicehovics: soo, the story is 

1. latest cordova cli (11) pins cordova-android platform to 10.1.2 https://cordova.apache.org/news/2021/12/21/cordova-cli-11.0.0-release.html
2. there is cordova-android 11 released on 12 of july 2022. You can always update platform by doing `cordova platform update android@11.0.0`, but it's not the default
3. cordova-android 11.* has following gradle defaults. https://github.com/apache/cordova-android/blob/master/framework/cdv-gradle-config-defaults.json . Cordova-android 10.* has SDK_VERSION set to 30. This translates to compileSdkVersion:30
4. work-runtime > "2.7.0-alpha04" has compileSdkVersion:31.
5. This produces following exception

```
 The minCompileSdk (31) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-30).
     Dependency: androidx.work:work-runtime:2.7.0
```

so, eventually this version pinning can be removed, but think not yet? Or new projects would have the error until android platform is upgraded

plugins/src/android/OptimoveInitProvider.java
vvoicehovics: k, updated

plugins/src/android/optimove.gradle
vvoicehovics: to not depend on an alpha dependency new plan is to update documentation asking to add 

```
<preference name="android-targetSdkVersion" value="31" />
```

to config.xml

PR #20, opened 2022-08-16
Cordova swift branch messaging

BBHOptimove:
Add messaging to Cordova SDK. This includes:
- push, inApp, inApp inbox APIs
- rename `Example_App` to remove underscore -- cannot have underscore in bundle identifiers
- more robust handing of push opens.


plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,25 +1,89 @@
...
+            pluginInstance.sendMessageToJs(type: "pushOpened", data: getPushNotificationMap(pushNotification: notification))
+        })
+
+        if #available(iOS 10, *) {
+            builder.setPushReceivedInForegroundHandler(pushReceivedInForegroundHandlerBlock: { notification, completionHanlder in
+                if (OptimoveSDKPlugin.optimovePluginInstance != nil){
+                    OptimoveSDKPlugin.optimovePluginInstance.sendMessageToJs(type: "pushReceived", data: getPushNotificationMap(pushNotification: notification))
+                }
+
+                completionHanlder(UNNotificationPresentationOptions.alert)
```
cgwyllie, line 61: Typo in name

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,25 +1,89 @@
...
+            })
+        }
+
+        if (configValues[inAppConsentStrategy] != InAppConsentStrategy.disabled.rawValue) {
+            builder.enableInAppMessaging(inAppConsentStrategy: configValues[inAppConsentStrategy] == InAppConsentStrategy.autoEnroll.rawValue ? OptimoveSDK.InAppConsentStrategy.autoEnroll : OptimoveSDK.InAppConsentStrategy.explicitByUser)
```
cgwyllie, line 66: Prefer explicit coverage of cases / no 'default' else? (maybe acceptable as unlikely changing options too soon)

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -1,25 +1,89 @@
...
         }
 
+        builder.setInAppDeepLinkHandler(inAppDeepLinkHandlerBlock: { data in
+            if (OptimoveSDKPlugin.optimovePluginInstance != nil){
+                OptimoveSDKPlugin.optimovePluginInstance.sendMessageToJs(type: "inAppDeepLink", data: getInappButtonPressMap(inAppButtonPress: data))
```
cgwyllie, line 71: Can simplify w/ optional chaining?

i.e. OptimoveSDKPlugin.optimovePluginInstance?.send...

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -139,4 +193,266 @@ import NotificationCenter
...
+            //return
+        }
+
+        let pluginResult:CDVPluginResult = CDVPluginResult.init(status: CDVCommandStatus_OK)
+        pluginResult.setKeepCallbackAs(true)
+        self.commandDelegate.send(pluginResult, callbackId: command.callbackId)
+    }
+
+    @objc(clearContext:)
+    func clearContext(command: CDVInvokedUrlCommand){
```
cgwyllie, line 222: When is this used?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -139,4 +193,266 @@ import NotificationCenter
...
+        if (OptimoveSDKPlugin.cordovaCommand == nil){
+            OptimoveSDKPlugin.cordovaCommand = command
+        }
+
+        if let notification = OptimoveSDKPlugin.pendingPush {
+            OptimoveSDKPlugin.optimovePluginInstance.sendMessageToJs(type: "pushOpened", data: OptimoveSDKPlugin.getPushNotificationMap(pushNotification: notification))
+            OptimoveSDKPlugin.pendingPush = nil
+        }
+        else{
+            let pluginResult: CDVPluginResult = CDVPluginResult(status: CDVCommandStatus_OK)
```
cgwyllie, line 239: Use of enum `.ok` vs `CDVCommandStatus_OK` ?

plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -139,4 +193,266 @@ import NotificationCenter
...
+            var pluginResult = CDVPluginResult.init(status: CDVCommandStatus_ERROR, messageAs: "Message not found or not available")
+            for msg in inboxItems {
+                if (msg.id != messageId){
+                    continue
+                }
+
+                let result = OptimoveInApp.presentInboxMessage(item: msg)
+                if (result == .PRESENTED){
+                    pluginResult = CDVPluginResult.init(status: CDVCommandStatus_OK)
+                }
```
cgwyllie, line 333: Is there any else to reject? Not sure what Android side of plugin did in this case.

cgwyllie: Ah nvm, see handled by default above (Y)

plugins/src/ios/OptimoveSDKPlugin.swift
vvoicehovics: prob was copied from Flutter, changed to switch

plugins/src/ios/OptimoveSDKPlugin.swift
vvoicehovics: it's about cordova lifecycle, this one https://cordova.apache.org/docs/en/3.3.0/cordova/events/events.pause.html

I think on android this was necessary because when bg app native->js communication stops and we should be saving pending things

cgwyllie (approved):

PR #23, opened 2022-09-13
Swift error handling

vvoicehovics:
1. Fix inject-config ios errors. Create plist resource file when plugin added
2. Fix build errors. Generated swift header file is named `ProductModuleName-Swift.h`. Cordova plugin doesn't create a separate module, so, ProductModuleName is not known by the plugin. To satisfy compiler partly copy generated interface into `InitializePluginManager.m`
3. handle optional optimove/optimobile credentials


plugins/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -20,7 +20,26 @@ import NotificationCenter
...
+            if let val = configValues[optimoveMobileCredentialsKey] {
+            if (!val.isEmpty){
+                optimobileCredentials = val;
+            }
+        }
+
+        if (optimoveCredentials == nil && optimobileCredentials == nil) {
+            print("Invalid credentials! Please provide at least one set of credentials")
+            return nil;
+        }
```
cgwyllie, line 40: Do we need this guard, I think the native SDK would also throw/crash/abort if both sets are nil.

Or do we want to support this in a 'warning' fashion, i.e. can inject empty creds is a valid use case?

plugins/src/ios/Bridging-Header.h
```diff
@@ -26,4 +26,4 @@
 //
 
 #import <Cordova/CDV.h>
-#import "InitializePluginManager.h"
+//#import "InitializePluginManager.h"
```
cgwyllie, line 29: Remove commented code?

plugins/src/ios/OptimoveSDKPlugin.swift
vvoicehovics: ah, right, i wanted to crash, but cannot throw from swift's variable initializer. Native crashing is good for me

cgwyllie (approved):

PR #24, opened 2022-09-15
Cordova error handling js

TomBarzilay1:


Example_App/www/js/index.js
```diff
@@ -155,18 +155,18 @@ function pushRegister() {
 }
 document
   .getElementById("in-app-consent-true")
-  .addEventListener("click", inAppUpdateConsentTrue);
+  .addEventListener("click", updateConsentForUserTrue);
```
vvoicehovics, line 158: is this renaming good? 
in optimove sdk we have `OptimoveInApp.getInstance().updateConsentForUser(consented);`, but there it's clear that it's about inApp from `OptimoveInApp` 'context'

`updateConsentForUser` doesnt tell what this consent is about

plugins/www/Optimove.js
```diff
@@ -77,139 +99,199 @@ function nativeMessageHandler(message) {
...
+            ]);
+        });
+    },
+
+    registerUser: function (userId, userEmail) {
+        return new Promise((resolve, reject) => {
+            var invalidRegister = false;
+            if (!isNonEmptyString(userId)) {
+                reject("Invalid user id");
+                invalidRegister = true;
```
vvoicehovics, line 163: better to return here? Once it's rejected it stays rejected, and (maybe) message doesnt change

plugins/www/Optimove.js
```diff
@@ -77,139 +99,199 @@ function nativeMessageHandler(message) {
...
+
+    pushRegister: function () {
+        return new Promise((resolve, reject) => {
+            exec(resolve, reject, "OptimoveSDKPlugin", "pushRegister", []);
+        });
+    },
+
+    updateConsentForUser: function (consented) {
+        return new Promise((resolve, reject) => {
+            if (typeof consented !== "boolean") {
```
vvoicehovics, line 194: can move into a function like other

plugins/www/Optimove.js
```diff
@@ -11,6 +11,28 @@ document.addEventListener("deviceready", init, false);
...
+function isNonEmptyString(val) {
+  return val && isString(val);
+}
+
+function isFunction(param) {
+  return typeof param === "function";
+}
+
+function isInboxItemValid(inboxItem) {
+  var mandatoryValues = ["id", "title", "subtitle", "sentAt"];
```
vvoicehovics, line 27: why specifically these fields? inbox item has more, but the only required is id. + with these fields example app will stop working as in `getInboxItemForTesting` we only give id to a fake inbox item.

Will this become redundant when we convert it to typescript? So, perhaps 
1. remove this validation function
2. convert index.js and Optimove.js to typescript
3. update `getInboxItemForTesting` to match typescript interface

plugins/www/Optimove.js
```diff
@@ -11,6 +11,28 @@ document.addEventListener("deviceready", init, false);
 document.addEventListener("resume", resume, false);
 document.addEventListener("pause", pause, false);
 
+function isString(val) {
```
vvoicehovics, line 14: if we convert this to typescript many of type checks will become redundant. But still can keep checking for "", for example

plugins/www/Optimove.js
```diff
@@ -1,4 +1,4 @@
-var exec = require("cordova/exec");
+const { isFunction } = require("util");
```
vvoicehovics, line 1: what is this? we define `isFunction` ourselves here

plugins/www/Optimove.js
```diff
@@ -77,139 +99,199 @@ function nativeMessageHandler(message) {
...
+                reject,
+                "OptimoveSDKPlugin",
+                "inAppDeleteMessageFromInbox",
+                [inAppInboxItem.id]
+            );
+        });
+    },
+
+    setOnInboxUpdatedHandler: function (handler) {
+        if (!isFunction(handler)) {
```
vvoicehovics, line 264: this can be null

PR #26, opened 2022-09-18
Convert plugin javascript code to typescript

TomBarzilay1:
This PR converts plugin js code to typescript.

1. `cd project/plugins; npm run build` to generate type declarations and minified js bundle
2. Added interfaces for all models accessible via API
3. Align data returned by native platforms to js. Aligned with Flutter SDK

TO COME: convert example app to ts, verify all interfaces usable as expected.


plugins/package.json
```diff
@@ -25,6 +26,15 @@
     ]
   },
   "devDependencies": {
-    "cordova-ios": "^6.2.0"
+    "@babel/preset-typescript": "^7.18.6",
+    "babel": "^6.23.0",
+    "babel-core": "^6.26.3",
+    "babel-loader": "^7.1.5",
+    "cordova-ios": "^6.2.0",
+    "source-map-loader": "^4.0.0",
```
vvoicehovics, line 34: are all of the dependencies here used ? For example, `source-map-loader` or `babel-loader` are not in webpack config. Or they are required by other dependencies? Could you describe what are they for and how applied

plugins/src/core/ddl.ts
```diff
@@ -0,0 +1,10 @@
+interface DeepLinkContent {
+  title?: string;
+  description?: string;
+}
+
+export interface DeepLink {
+  content: DeepLinkContent;
+  data: any;
```
vvoicehovics, line 8: data is a json object, mb can be a bit more specific with type

plugins/package.json
TomBarzilay1: these dependencies were auto generated in one of the iterations of trying to make the bundling work, I will delete them

plugins/src/core/ddl.d.ts
```diff
@@ -0,0 +1,10 @@
+interface DeepLinkContent {
+    title?: string;
+    description?: string;
+}
+export interface DeepLink {
+    content: DeepLinkContent;
+    data: any;
+    url: string;
+}
+export {};
```
vvoicehovics, line 10: ?

plugins/src/core/index.ts
```diff
@@ -0,0 +1,369 @@
+import * as cordova from "cordova";
+import {
+  PushNotificationHandler,
+  InAppDeepLinkHandler,
+  InAppInboxUpdatedHandler,
+  DeepLinkHandler,
+} from "./handlers";
+
+export interface OptimoveConfig{
+  pushReceivedHandler?: PushNotificationHandler | null;
```
vvoicehovics, line 10: are these not optional? They are always null or a function

plugins/src/core/index.ts
```diff
@@ -0,0 +1,369 @@
...
+
+interface InAppInboxItem {
+  id: number;
+  title: string;
+  subtitle: string;
+  availableFrom: string | "";
+  availableTo: string | "";
+  dismissedAt: string | "";
+  isRead: boolean;
+  sentAt?: string;
```
vvoicehovics, line 121: it's not enough to copy from the old SDK, ideally need to verify that types in Optimove SDKs match . for example sentAt i think is not optional?

plugins/src/core/index.ts
```diff
@@ -0,0 +1,369 @@
...
+      nativeMessageHandler,
+      reject,
+      "OptimoveSDKPlugin",
+      "checkIfPendingPushExists",
+      []
+    );
+  });
+}
+
+function nativeMessageHandler(message) {
```
vvoicehovics, line 91: can add type for the arg

plugins/src/core/index.ts
```diff
@@ -0,0 +1,369 @@
...
+  if (
+    handlerName === "inAppInboxUpdatedHandler" &&
+    typeof currentConfig[handlerName] === "function"
+  ) {
+    currentConfig[handlerName]?.();
+
+    return;
+  }
+
+  if (typeof currentConfig[handlerName] === "function") {
```
vvoicehovics, line 106: can be replaced with `currentConfig[handlerName] !== null`. And same for all typeof

plugins/src/core/index.ts
```diff
@@ -0,0 +1,369 @@
...
+  [key: string]: any;
+}
+interface InAppInboxSummary {
+  totalCount: number;
+  unreadCount: number;
+}
+
+const Optimove = {
+  /**
+   * Sets the User ID of the current user and starts the {@code Visitor} to {@code Customer} conversion flow.
```
vvoicehovics, line 135: are these descriptions taken from somewhere? Do consumers of the SDK really need to know about `starts the {@code Visitor} to {@code Customer} conversion flow.`?

plugins/src/core/ddl.ts
TomBarzilay1: image
 I took this object as a reference, but of course when I look at the actual object that we send from the native function the objects are different, I'll fix this now.

plugins/src/core/index.ts
TomBarzilay1: I used Android SDK and Kumulos Cordova SDK as a reference

plugins/src/core/push.ts
```diff
@@ -0,0 +1,7 @@
+export interface PushNotification {
+  id: number;
+  title?: string | null;
+  message?: string | null;
+  data?: { [key: string]: any } | null;
```
vvoicehovics, line 5: same as above. consistency with cordova ios side/Flutter, verify it's optional VS null

plugins/src/core/ddl.d.ts
TomBarzilay1: this file  was created automatically after the bundling, I'm not sure why the export {}; is needed, seems weird, I'll delete that.

plugins/src/core/index.ts
vvoicehovics: y, but java annotations dont belong to typescript

plugins/src/core/index.ts
TomBarzilay1: changed the handlers to non - optional

plugins/src/core/index.ts
TomBarzilay1: I aligned the interface with flutter sdk

plugins/src/core/handlers.ts
```diff
@@ -0,0 +1,17 @@
+import { PushNotification } from "./push";
+import { DeepLink } from "./ddl";
+export interface PushNotificationHandler {
+  (notification: PushNotification): void;
+}
+
+export interface InAppDeepLinkHandler {
+  (data: { [key: string]: any }): void;
```
vvoicehovics, line 8: InAppButtonPress is the argument, which is 
```
public struct InAppButtonPress {
    public let deepLinkData : [AnyHashable:Any]
    public let messageId : Int64
    public let messageData : NSDictionary?

    init(deepLinkData:[AnyHashable:Any], messageId:Int64, messageData:NSDictionary?) {
        self.deepLinkData = deepLinkData
        self.messageId = messageId
        self.messageData = messageData
    }
}
```

in my test i've received this `{"messageData":null,"deepLinkData":{"hellow":"win"},"messageId":119}`

Example_App/www/js/index.js
```diff
@@ -249,12 +247,4 @@ function inAppDeleteMessageFromInbox() {
     error
   );
 }
-document
```
vvoicehovics, line 252: removed listener?

Example_App/www/js/index.js
```diff
@@ -249,12 +247,4 @@ function inAppDeleteMessageFromInbox() {
...
-  .getElementById("get-visitor-id-button")
-  .addEventListener("click", getVisitorId);
```
vvoicehovics, line 254: removed listener?

plugins/src/core/ddl.ts
vvoicehovics: resolution can be an enum

plugins/src/core/index.ts
vvoicehovics: aaaa, this is not `imagePath`, it's `imageUrl`. and availableTo/From are "", not null. Can double check this again?

plugins/src/core/ddl.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+ 
+export interface DeepLink {
+  link: string;
+  resolution: DeepLinkResolution;
+  data: DeepLinkData;
+}
+
+export enum DeepLinkResolution {
+  LOOKUP_FAILED,
```
vvoicehovics, line 18: enum values here are strings, not integers

plugins/src/core/inApp.ts
```diff
@@ -0,0 +1,23 @@
+export interface InAppInboxItem {
+  id: number;
+  title: string;
+  subtitle: string;
+  availableFrom: string | null; // Date?
+  availableTo: string | null;
+  dismissedAt: string | null;
```
vvoicehovics, line 7: availableFrom, availableTo and dismissedAt are being passed as "" from native. nulls are what Flutter does. Can we align native with js? you can add/perhaps even test in android. ill add in swift

plugins/src/core/index.ts
vvoicehovics: same for @link

plugins/src/core/push.ts
```diff
@@ -0,0 +1,7 @@
+export interface PushNotification {
+  id: number;
+  title: string | null;
+  message: string | null;
+  data: { [key: string]: any } | null;
```
vvoicehovics, line 5: can we use `JSON | null` like in other places

vvoicehovics: replaced JSON with `{ [key: string]: any }`

plugins/src/core/ddl.ts
```diff
@@ -0,0 +1,23 @@
+export interface DeepLinkContent {
+  title?: string;
+  description?: string;
+}
+export interface DeepLinkData {
+  data: { [key: string]: any };
```
cgwyllie, line 6: Can possibly use `Record<string,any>` for these types of declarations?

plugins/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+        "target": "es2015",
+        "allowSyntheticDefaultImports": true,
+        "declaration": true,
+        "outDir": "./src/core/declarations",
+        "emitDecoratorMetadata": true,
+        "experimentalDecorators": true,
```
cgwyllie, line 8: Is this 'standard'?

plugins/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+        "rootDir": "./src/core",
+        "module": "esnext",
+        "moduleResolution": "node"
+    },
+     "include":[
+        "src/core/*.ts"
```
cgwyllie, line 14: Can/should we glob files to allow future module additions more easily?

e.g. `src/**/*.ts`

plugins/package.json
```diff
@@ -5,6 +5,7 @@
   "description": "A sample Apache Cordova application that responds to the deviceready event.",
   "main": "index.js",
```
cgwyllie, line 6: I think we can/should perhaps declare the "types" field to point at index.d.ts for the module?

Maybe it's automatically picked up based on some convention, but if in a non-standard location I think we have to declare it.

plugins/plugin.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version='1.0' encoding='utf-8'?>
 <plugin id="optimove-cordova-plugin" version="1.0.0" xmlns="http://apache.org/cordova/ns/plugins/1.0" xmlns:android="http://schemas.android.com/apk/res/android">
     <name>OptimoveSDKPlugin</name>
+    <js-module name="OptimoveCore" src="www/OptimoveCore.js"/>
     <js-module name="Optimove" src="www/Optimove.js">
```
cgwyllie, line 5: Do we need to declare both modules / what's the reason for the split?

cgwyllie: (ah sorry see this is how we have it set up in Kumulos plugin too)

plugins/src/core/ddl.ts
vvoicehovics: updated. what's the benefit, only slightly neater syntax?

plugins/plugin.xml
vvoicehovics: i think it's possible to output a module from webpack immediately https://webpack.js.org/configuration/output/#outputmodule . But it's an experimental feature and i figured easier to not mess up with it

plugins/package.json
vvoicehovics: yeah,  "main" is quite certainly incorrect. Im not sure if my change is 100% right, in the old SDK we used to copy some of the types into root folder for some reason.

Maybe can verify types are found when typescriptify the app

plugins/src/core/ddl.ts
cgwyllie: Yeah, just a nice utility type and a bit easier to type out (I think).

https://www.typescriptlang.org/docs/handbook/utility-types.html

plugins/package.json
cgwyllie: TS type declarations should also help (if they are found) in non-TS apps the are edited in a TS def aware editor like VSCode. They typing info should be found in the JS app & help w/ arg hints etc. if it's working properly I think.

cgwyllie: Pick this up as follow-up.

cgwyllie (approved):

PR #35, opened 2022-10-06
Final snags for Cordova SDK

vvoicehovics:
1. Publish Cordova SDK as a scoped npm package
2. push uncommitted base ios version update
3. bugfix: incorrect casing breaks `getVisitorId` on ios
4. in ExampleApp add label when printing 'pushReceivedHandler' payload (just like for other handlers)


ExampleApp/package.json
```diff
@@ -28,7 +28,7 @@
       "android"
     ],
     "plugins": {
-      "optimove-cordova-plugin": {}
+      "cordova-sdk": {}
```
cgwyllie, line 31: With it being scoped, is there any risk of collisions I wonder.. Can/should we prefix in the 'id', but not 'name' fields of meta-data? (if this is even possible..)

vvoicehovics: suppose no reason not to, updated

cgwyllie (approved):

PR #40, opened 2023-01-27
Ionic support + API updates

vvoicehovics:
### Description of Changes

1. Remove irrelevant manifest modifications (apparently inherited from the early sdk versions)
2. minor tweaks to types in the ExampleApp. ExampleApp works with cordova, not ionic
3. add signOutUser and pushUnregister API. Add to ExampleApp
4. update android base sdk `6.* --> 7.0.1`, use new api for requesting push token (pushRegister --> pushRequestDeviceToken)
5. when presenting in-apps return OptimoveInAppPresentationResult similar to other sdks

### Breaking Changes

- pushRegister --> pushRequestDeviceToken
- presenting an in-app message always returns a resolved promise with OptimoveInAppPresentationResult 

### Release Checklist

Prepare:

-   [x] `cd cordova-sdk` and `npm run build` to produce minified artifacts
-   [x] `cd ExampleApp` and `npm run build` to update `index.js`
-   [x] Detail any breaking changes. Breaking changes require a new major version number
-   [x] Update type declarations in `cordova-sdk/index.d.ts`

Bump versions in:

-   [x] package.json
-   [x] plugin.xml
-   [x] src/ios/OptimoveSDKPlugin.swift
-   [x] src/android/OptimoveInitProvider.java

Release:

-   [ ] Squash and merge to master
-   [ ] Create tag from master matching chosen version

Update wiki:


cordova-sdk/package.json
```diff
@@ -25,6 +25,7 @@
     ]
   },
   "devDependencies": {
+    "@optimove-inc/cordova-sdk": "file:../cordova-sdk",
```
cgwyllie, line 28: What's the meaning to depend on the SDK in the SDK? (would expect in the sample app, but here seems strange, maybe it's correct, just don't understand)

cordova-sdk/plugin.xml
```diff
@@ -53,16 +53,7 @@
             <!-- Optimove initProvider-->
             <provider android:authorities="${applicationId}.optimoveinitprovider" android:name="com.optimove.android.cordova.OptimoveInitProvider" android:exported="false" />
         </config-file>
-        <edit-config file="AndroidManifest.xml" target="/manifest" mode="merge">
-            <manifest xmlns:tools="http://schemas.android.com/tools" />
```
cgwyllie, line 57: We don't use `tools:` nodes for anything else? (I think no, just confirming)

cgwyllie (approved):
LGTM

cordova-sdk/package.json
vvoicehovics: ofc it's not correct, removed 😬

cordova-sdk/plugin.xml
vvoicehovics: let's start by taking down edit-config removing backup rules

PR #49, opened 2023-12-10
rc: 2.2.0

eligutovsky:
### Description of Changes

- Update `OptimoveSDK` dependency to `5.6.0`
- Update `cordova-ios` to `7.0.1`
- Update deployment target and swift version

### Breaking Changes

-   None

### Release Checklist

Prepare:

-   [ ] `cd cordova-sdk` and `npm run build` to produce minified artifacts
-   [ ] `cd ExampleApp` and `npm run build` to update `index.js`
-   [ ] Detail any breaking changes. Breaking changes require a new major version number
-   [ ] Update type declarations in `cordova-sdk/index.d.ts`

Bump versions in:

-   [x] package.json
-   [x] plugin.xml
-   [x] src/ios/OptimoveSDKPlugin.swift
-   [ ] src/android/OptimoveInitProvider.java

Release:

-   [ ] Squash and merge to main
-   [ ] Create tag from main matching chosen version

Update wiki:


cordova-sdk/plugin.xml
```diff
@@ -1,32 +1,35 @@
...
         </config-file>
+        <resource-file src="scripts/optimove.plist" target="optimove.plist" />
         <source-file src="src/ios/OptimoveSDKPlugin.swift" />
         <header-file src="src/ios/InitializePluginManager.h" />
         <source-file src="src/ios/InitializePluginManager.m" />
         <header-file src="src/ios/Bridging-Header.h" type="BridgingHeader" />
         <podspec>
             <config>
-                <source url="https://github.com/CocoaPods/Specs.git"/>
+                <source url="https://cdn.cocoapods.org/"/>
```
eligutovsky, line 25: this change speed up fetching cocoapods

cordova-sdk/plugin.xml
```diff
@@ -1,32 +1,35 @@
...
             </config>
             <pods>
-                <pod name="OptimoveCore" spec="5.1.1" />
-                <pod name="OptimoveSDK" spec="5.1.1" />
+                <pod name="OptimoveSDK" spec="~> 5" />
```
eligutovsky, line 28: Optimove SDK has an inner dependency on Optimove Core an will be installed in bulk.

vvoicehovics: i think intention in hybrid sdks was to pin to a specific base sdk version. it means
- bugs in base sdks dont propagate to each hybrid
- easier to relate issues to a specific version

eligutovsky: Ok. Let's pin the latest known version

cordova-sdk/plugin.xml
```diff
@@ -1,32 +1,35 @@
...
     <hook type="before_prepare" src="scripts/inject-config.js" />
+
     <platform name="ios">
-        <resource-file src="scripts/optimove.plist" target="optimove.plist" />
-        <config-file target="config.xml" parent="/*">
+        <config-file parent="/*" target="config.xml">
             <feature name="OptimoveSDKPlugin">
                 <param name="ios-package" value="Optimove_Cordova" />
             </feature>
+            <preference name="SwiftVersion" value="5" />
```
vvoicehovics, line 15: if by default it doesnt set swift version of the project. Does it mean setting it to 5 is what potentially could lead to conflicts? (especially if 6 is released 1 day)

eligutovsky: Cordova could operate with Swift version 4. It is good practice to set the language version.
When it will be Swift 6 -- we will change it here as well

vvoicehovics: would this force consumers of the plugin into swift 6 as well?

eligutovsky: > would this force consumers of the plugin into swift 6 as well?

Once we will stop support Swift 5, it is highly possible that it will force them to use Swift 6.

vvoicehovics: still dont undertsand, what happens if we set swift 6, other plugin sets swift 5? plugin added later wins?

eligutovsky: The build does not pass CI if delete this line:
https://github.com/optimove-tech/Optimove-SDK-Cordova/actions/runs/7195249630/job/19597497704#step:6:71

eligutovsky: > still dont undertsand, what happens if we set swift 6, other plugin sets swift 5? plugin added later wins?

We will see it in the future.

PR #51, opened 2024-09-09
AB#228381 - Add delayed initialisation

k-antipochkin:
### Description of Changes

(briefly outline the reason for changes, and describe what's been done)

### Breaking Changes

-   None

### Release Checklist

Prepare:

-   [ ] `cd cordova-sdk` and `npm run build` to produce minified artifacts
-   [ ] `cd ExampleApp` and `npm run build` to update `index.js`
-   [ ] Detail any breaking changes. Breaking changes require a new major version number
-   [ ] Update type declarations in `cordova-sdk/index.d.ts`

Bump versions in:

-   [ ] package.json
-   [ ] plugin.xml
-   [ ] src/ios/OptimoveSDKPlugin.swift
-   [ ] src/android/OptimoveInitProvider.java

Release:

-   [ ] Squash and merge to main
-   [ ] Create tag from main matching chosen version

Update wiki:


cordova-sdk/src/android/OptimoveInitProvider.java
```diff
@@ -38,24 +42,30 @@ public class OptimoveInitProvider extends ContentProvider {
...
+    }
 
-        OptimoveConfig.Builder configBuilder = new OptimoveConfig.Builder(optimoveCredentials, optimoveMobileCredentials);
-
-        if (optimoveMobileCredentials == null) {
-            Optimove.initialize(app, configBuilder.build());
-            return true;
-        }
-
+    if (optimoveMobileCredentials != null || enableDelayedInitialization) {
```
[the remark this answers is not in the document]
vvoicehovics: even similar to base sdks, depending on init type can 1st get feature set (from config or from presence of optimoveMobileCredentials) and then do in-app, ddl etc inits if set contains optimobile.

vvoicehovics:
perhaps optimove.json can get confusing if it is used for delayed and immediate init? 

1 other way is to have optimove.json and optimoveDelayed.json, then prepare script checks any one that exists, if both exist -- crash. is it worth or same ~stuff anyway ?

cordova-sdk/plugin.xml
```diff
@@ -22,8 +22,8 @@
                 <source url="https://github.com/CocoaPods/Specs.git"/>
             </config>
             <pods>
-                <pod name="OptimoveCore" spec="5.1.1" />
-                <pod name="OptimoveSDK" spec="5.1.1" />
+                <pod name="OptimoveCore" spec="6.0.1" />
```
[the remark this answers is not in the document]
dsellick-opti: makes sense will see about getting this released so we can include it in this PR.

ExampleApp/package.json
```diff
@@ -17,18 +17,23 @@
   "devDependencies": {
     "@optimove-inc/cordova-sdk": "file:../cordova-sdk",
     "cordova-android": "^11.0.0",
-    "cordova-ios": "^6.2.0",
+    "cordova-ios": "^7.1.1",
+    "ios": "^0.0.1",
```
vvoicehovics, line 21: what's this one?

vvoicehovics: also semver is not used?

cordova-sdk/scripts/inject-config.js
```diff
@@ -134,32 +136,38 @@ function prepareAndroid(context, config) {
...
+    DELAYED_INITIALIZATION_REGION: delayedInitialization && isString(delayedInitialization.region)
+        ? delayedInitialization.region
+        : "",
+    ANDROID_PUSH_NOTIFICATION_ICON_NAME:
+        android !== undefined &&
+        !isEmpty(android.pushNotificationIconName) &&
+        isString(android.pushNotificationIconName)
+            ? android.pushNotificationIconName
+            : "",
+    DELAYED_INITIALIZATION_ENABLE_OPTIMOVE: featureSet && featureSet.enableOptimove === true,
```
vvoicehovics, line 168: simpler to have in optimove config `featureSet: [optimove, optimobile]`? later can add pref center with changing less

ExampleApp/package.json
dsellick-opti: this will have been from removing and adding the platfom as part of development.  works with the latest version I will test with 6.2.0 as well.  semver is a dependacy pulled in from the cordova package and its dependacies

cordova-sdk/src/android/OptimoveInitProvider.java
```diff
@@ -73,20 +93,23 @@ public boolean onCreate() {
             final int iconResource = resources.getIdentifier(pushNotificationIconName, "drawable", packageName);
             configBuilder.setPushSmallIconId(iconResource);
         }
+    }
 
-        overrideInstallInfo(configBuilder);
+    overrideInstallInfo(configBuilder);
```
vvoicehovics, line 98: this stuff and below also only relevant for optimobile. e.g. previously we had

```
    if (optimoveMobileCredentials == null) {
            Optimove.initialize(app, configBuilder.build());
            return true;
        }
         ..... all other stuff here, setPushActionHandler, setOnInboxUpdated, setDeepLinkHandler, overrideInstallInfo etc
      

```

cordova-sdk/src/ios/OptimoveSDKPlugin.swift
```diff
@@ -158,13 +162,63 @@ enum InAppConsentStrategy: String {
             }
         }
 
+        if let enableDelayedInitialization = configValues[delayedInitializationEnable],
```
vvoicehovics, line 165: silly github doesnt lt commenting on unchanged line

here `if (configValues[optimoveMobileCredentialsKey] == nil) {`

should have same logic as in android. Currently if there is no optimoveMobileCredentialsKey (e.g. for any delayed init), we wont do optimobile initialisation below

ExampleApp/package.json
vvoicehovics: smth seems not right, what is this package https://www.npmjs.com/package/ios
- published last time 10 y ago
- version 0.0.1
- keywords "hello world"
- doesnt have any description

to me seems like a random test package with xiaosheng trying out stuff, but having 1000s of downloads because of name "ios" :D

transitive dependencies shouldnt be added here by cordova

vvoicehovics: some other packages he has

image

kris-opti (approved):

